PR #1, opened 2019-04-05
Initial client work

docmerlin:


client.go
```diff
@@ -0,0 +1,312 @@
...
+		return dur, "", err
+	}
+
+	// we shouldn't see this
+	if resp.StatusCode != http.StatusNoContent && resp.StatusCode != http.StatusOK {
+		var err = errors.New(string(body))
+		return 0, "", err
+	}
+
+	version := resp.Header.Get("X-Influxdb-Version")
```
goller: Is this still part of the API?  If not, I think it would be nice to add as an issue in influxdb.

client.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	req.Header.Set("Content-Type", "text/plain; charset=utf-8")
+
+	if c.contentEncoding == "gzip" {
+		req.Header.Set("Content-Encoding", "gzip")
+	}
+
+	req.Header.Set("User-Agent", c.userAgent)
+	req.Header.Set("Authorization", c.authorization)
```
goller: I think this needs to add the "Bearer " + c.authorization

docmerlin: I think we are using "Token " + "token string."

The `WithToken(string) Option` function handles that.

client.go
```diff
@@ -0,0 +1,312 @@
...
+		err = &genericRespError{
+			Code:    resp.Status,
+			Message: "service is temporarily unavaliable",
+		}
+		if err2 := c.backoff(triesPtr, resp, err); err2 != nil {
+			return err2
+		}
+		goto doRequest
+	// split up entities that are just too big
+	case http.StatusRequestEntityTooLarge:
```
danielnelson: Return error and let the user send less.  Hiding this could make the client needlessly inefficient, its complicated, and its likely it won't even work.  The last time I saw this it was a single field key causing the problem.

clientoptions.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// UserAgent is the http User Agent, defaults to "InfluxDBClient" plus os and version info.
+	UserAgent string
+
+	// Timeout for influxdb writes, if set to zero, it defaults to a 20 second timeout. This is a difference from the influxdb1-client.
+	Timeout time.Duration
+
+	// InsecureSkipVerify gets passed to the http client, if true, it will
+	// skip https certificate verification. Defaults to false.
+	InsecureSkipVerify bool
```
danielnelson: Just make this part of the TLSConfig, then you won't need to worry about their conflicting interaction.

clientoptions.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// TLSConfig allows the user to set their own TLS config for the HTTP
+	// Client. If set, this option overrides InsecureSkipVerify.
+	TLSConfig *tls.Config
+
+	// Proxy configures the Proxy function on the HTTP client.
+	Proxy func(req *http.Request) (*url.URL, error)
+}
+
+// WithV1Config is an option for setting config in a way that makes it easy to convert from the old config.
```
danielnelson: What old config?

client.go
```diff
@@ -0,0 +1,312 @@
...
+	for i := range m {
+		if _, err = e.Encode(m[i]); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// backoff is a helper method for backoff, triesPtr must not be nil.
```
jacobmarble: > backoff is a helper method for backoff

Does not make sense.

clientoptions.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		if conf.InsecureSkipVerify || conf.TLSConfig != nil || conf.Proxy != nil {
+			panic("Unimplemented")
+		}
+		return nil
+	}
+}
+
+// WithNoCompression returns an option for writing the data to influxdb without compression.
+func WithNoCompression() Option {
```
danielnelson: This option conflicts with WithGZIP, what if you pass both of these?  Maybe have a single WithCompression() option?

clientoptions.go
```diff
@@ -0,0 +1,150 @@
...
+// WithToken returns an option for setting a token.
+func WithToken(token string) Option {
+	return func(c *Client) error {
+		c.authorization = "Token " + token
+		return nil
+	}
+}
+
+// WithNoFailOnFieldError returns an option that changes the client so that Client.Write will not fail when there is an encoding field error.
+func WithNoFailOnFieldError() Option {
```
danielnelson: `WithIgnoreFieldError`?

client.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	version := resp.Header.Get("X-Influxdb-Version")
+	return dur, version, nil
+}
+
+// Write writes metrics to a bucket, and org.  It retries intelligently.
+// If the write is too big, it retries again, after breaing the payloads into two requests.
+func (c *Client) Write(ctx context.Context, bucket, org string, m ...Metric) (err error) {
+	if c.authorization == "" {
```
danielnelson: This is just me rambling about what I'm thinking, not a real comment...

If something is required, it shouldn't be an option.  A good API is hard to use incorrectly, but the Setup stuff, and even an unauthenticated endpoint like Ping, kind of throws a wrench in this.  What to do... I'm not sure, but I'm wondering do we really need Setup at all, and if so do we need a different client for it.

Maybe two types and with an upgrade procedure: Client() -> AuthenticatedClient

models.go
```diff
@@ -0,0 +1,145 @@
...
+	return m.Fields
+}
+
+// Time is the timestamp of a metric.
+func (m *RowMetric) Time() time.Time {
+	return m.TS
+}
+
+// SortTags orders the tags of a metric alphnumerically by key.
+func (m *RowMetric) SortTags() {
```
danielnelson: Tags should probably always be sorted, but I wouldn't add a function.

models.go
```diff
@@ -0,0 +1,145 @@
...
+// that has methods to make it easy to add tags and fields
+type RowMetric struct {
+	NameStr string
+	Tags    []*lp.Tag
+	Fields  []*lp.Field
+	TS      time.Time
+}
+
+// TagList returns a slice containing Tags of a Metric.
+func (m *RowMetric) TagList() []*lp.Tag {
```
danielnelson: Might be worth attempting to prevent the user from directly modifying these items, by returning an interface.  This is something I've been meaning to do in Telegraf.

client.go
```diff
@@ -0,0 +1,322 @@
...
+		}
+	}
+
+	return nil
+}
+
+// backoff is a helper method for backoff, triesPtr must not be nil.
+func (c *Client) backoff(triesPtr *uint64, resp *http.Response, err error) error {
+	tries := atomic.LoadUint64(triesPtr)
+	if c.maxRetries >= 0 || int(tries) >= c.maxRetries {
```
sebito91: Why would we bail if the user has set this to be .> 0? Did you mean to make this `c.maxRetries == 0`?

client.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+	tries := uint64(0)
+	return c.write(ctx, bucket, org, &tries, m...)
+}
+
+func (c *Client) write(ctx context.Context, bucket, org string, triesPtr *uint64, m ...Metric) error {
+	//r, w := io.Pipe()
+	buf := &bytes.Buffer{}
+	e := lp.NewEncoder(buf)
+doRequest:
```
sebito91: I would love to see a refactor of this to remove the `goto` dependency. I'm not convinced we need that here when a loop could suffice and clarify things dramatically.

client.go
```diff
@@ -0,0 +1,322 @@
...
+	if c.maxRetries >= 0 || int(tries) >= c.maxRetries {
+		return maxRetriesExceededError{
+			err:   err,
+			tries: c.maxRetries,
+		}
+	}
+	retry := 0
+	if resp != nil {
+		retryAfter := resp.Header.Get("Retry-After")
+		retry, _ = strconv.Atoi(retryAfter) // we ignore the error here because an error already means retry is 0.
```
sebito91: You shouldn't ignore this error, might be useful for us to check on.

client.go
```diff
@@ -0,0 +1,322 @@
...
+	case "unix":
+	default:
+		return "", fmt.Errorf("unsupported scheme: %q", u.Scheme)
+	}
+	u.RawQuery = params.Encode()
+	return u.String(), nil
+}
+
+// Close closes any idle connections on the Client.
+func (c *Client) Close() error {
```
sebito91: Don't think this is actually used anywhere?

errors.go
```diff
@@ -0,0 +1,63 @@
...
+		return coalescingError(err[1:])
+	}
+	if len(err) == 1 {
+		return err[0]
+	}
+	return nil
+}
+
+type genericRespError struct {
+	Code      string
```
sebito91: Are we sure this is `Code` or the `Status`? In the usage in your client you assign the `resp.Status` to this field instead of the `resp.StatusCode`. I would make them consistent for clarity.

models.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// Name returns the name of the metric.
+func (m *RowMetric) Name() string {
+	return m.NameStr
+}
+
+func convertField(v interface{}) interface{} {
+	switch v := v.(type) {
+	case float64:
```
sebito91: Just combine these...

```
switch v := v.(type) {
case float64, int64, string, bool:
    return v
case ...:
}
```

query.go
```diff
@@ -0,0 +1,75 @@
...
+	cleanup := func() {
+		resp.Body.Close()
+	}
+	defer func() { cleanup() }()
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		r := io.LimitReader(resp.Body, 1<<14) // only support errors that are 16kB long, more than that and something is probably wrong.
```
sebito91: I'm confused why this is needed, what are we preventing against here?

setup.go
```diff
@@ -0,0 +1,132 @@
...
+
+	setupResult := &SetupResult{}
+	if err := json.NewDecoder(resp.Body).Decode(setupResult); err != nil {
+		return nil, err
+	}
+	if err != nil {
+		return nil, err
+	}
+	if setupResult.Code != "conflict" && resp.StatusCode == http.StatusCreated && setupResult.Auth.Token != "" {
+		c.l.Lock()
```
sebito91: I'm not sure I follow why the lock is required here? Is this necessary?

useragent.go
```diff
@@ -0,0 +1,10 @@
+package influxdb
+
+import (
+	"fmt"
+	"runtime"
+)
+
+func ua() string {
+	return fmt.Sprintf("InfluxDBClient/v0.0.1  (%s; %s; %s)", "golang", runtime.GOOS, runtime.GOARCH)
```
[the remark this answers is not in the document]
docmerlin: That is something I am working on and will be a separate PR.

setup.go
```diff
@@ -0,0 +1,132 @@
...
+	"encoding/json"
+	"errors"
+	"net/http"
+)
+
+// Setup sets up a new influxdb server.
+// It requires a client be set up with a username and password.
+// If successful will add a token to the client.
+// RetentionPeriodHrs of zero will result in infinite retention.
+func (c *Client) Setup(ctx context.Context, bucket, org string, RetentionPeriodHrs int) (*SetupResult, error) {
```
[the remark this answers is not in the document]
docmerlin: normally a time.Duration would make sense, but I switched too an int, because the API expects an hours int and I do not want to imply that the period would be more fine grained than hours.

setup.go
```diff
@@ -0,0 +1,132 @@
...
+	if c.username == "" || c.password == "" {
```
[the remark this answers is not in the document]
docmerlin: They would also be used once basic auth style connections are accepted in influx, otherwise I would agree with you.

README.md
```diff
@@ -1,6 +1,39 @@
...
+
+	// The actual write...
+	if err := influx.Write(context.Background(), "my-awesome-bucket", "my-very-awesome-org", myMetrics...); err != nil {
+		fmt.Println(err) // as above use your own error handling here.
+	}
+	influx.Close()
+```
+
+## Releases
+We will be using git-flow style releases, the current stable release will be listed in the master readme.
```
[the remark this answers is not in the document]
docmerlin: No, it is our intent to tag and have the highest tag also be master.

http.go
```diff
@@ -0,0 +1,38 @@
...
+	"net/http"
+	"time"
+)
+
+func netTransport() *http.Transport {
+	return &http.Transport{
+		Dial: (&net.Dialer{
+			Timeout: 5 * time.Second,
+		}).Dial,
+		TLSHandshakeTimeout: 5 * time.Second,
```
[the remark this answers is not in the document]
docmerlin: they timeout different things.  the timeout on HTTP client is for the entire transfer.  For large files on slow connections you want a longer timeout client timeout.  For the dial and tls handshake you want a smaller one, generally.

http.go
```diff
@@ -0,0 +1,38 @@
...
+
+func defaultHTTPClient() *http.Client {
+	return &http.Client{
+		Timeout:   time.Second * 20,
+		Transport: netTransport(),
+	}
+}
+
+// HTTPClientWithTLSConfig returns an *http.Client with sane timeouts and the provided TLSClientConfig.
+func HTTPClientWithTLSConfig(conf *tls.Config) *http.Client {
```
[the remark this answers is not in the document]
docmerlin: It is not a test function.  It allows people to set up their own TLS client if they need custom certs etc.

client.go
```diff
@@ -0,0 +1,312 @@
...
+	contentEncoding  string
+	compressionLevel int
+	url              *url.URL
+	password         string
+	username         string
+	l                sync.Mutex
+	maxRetries       int
+	errOnFieldErr    bool
+	userAgent        string
+	authorization    string
```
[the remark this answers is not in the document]
docmerlin: It isn't a token,  It is the Authorization header.  I will add a comment.

client.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		return err
+	}
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	throwAway := make([]byte, 128)
+	for err == nil {
+		_, err = resp.Body.Read(throwAway)
```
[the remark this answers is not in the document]
docmerlin: The purpose was to full read the buffer if there were extraneous bytes, so that a connection could be reused, I've modified and moved it.

clientoptions.go
docmerlin: This (and that entire struct) is that way for compatibility with the old client.  There are better ways if you want a new client.

clientoptions.go
docmerlin: The config for the old golang client for influxdb1x.  I will make it more clear in the comment.

clientoptions.go
docmerlin: If you pass both, the later one overrides. I will make that clear in the comments for the `Option` type

models.go
docmerlin: It is a helper so someone making RowMetrics manually can sort their tags easily.  Will add that in the comment.

models.go
docmerlin: I *really* don't like that.  Also it won't stop a user from doing it, they end up just using reflection if they really want to.

client.go
docmerlin: I prefer the goto for this use case, its clearer almost every case involves breaking out of the loop.  The goto makes it clear that it is the unusual case where we retry.

client.go
docmerlin: I disagree.  If the header is non-numeric it means something bad (expected) happened, but we can still continue normally, there is no reason to stop execution or retry as a result.

client.go
docmerlin: It is used.

errors.go
docmerlin: Yes it is Code and a string.

The db returns an error in the form of a json string.
Example:
```json
{"code":"unauthorized","message":"unauthorized access"}
```

query.go
docmerlin: You want to flush the tcp buffer before closing the connection in order to reuse the connection, but if it is too full, it is faster to just make a new connection and let this one time out.

setup.go
docmerlin: To prevent a "harmless" data race.

models.go
danielnelson: I don't think this should be something the user of the library needs to concern themselves with, we should just do it under the hood.  Is there a reason the client user would want to sort outside of the encoding step?

models.go
danielnelson: It's more about hinting to the user that they shouldn't modify this without using the modification functions.  What don't you like about it?

clientoptions.go
danielnelson: While a comment is good, its always best to make an interface harder to use wrong.  Maybe this option could have an argument to give it more configurability without the possibility of it being used illegally.  I realize there are still ways it can be used wrong, such as sending the option twice, but I still think it would be helpful.

http.go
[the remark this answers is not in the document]
danielnelson: Okay then I don't think this function is needed, just let the user create the http.Client on their own.

client.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+	switch resp.StatusCode {
+	case http.StatusOK, http.StatusNoContent:
+	case http.StatusNotFound:
+		return &genericRespError{
+			Code:    resp.Status,
```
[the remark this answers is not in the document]
docmerlin: No, it is a status message , but called code by the error the db returns.

models.go
docmerlin: 1) Generally returning interfaces is awkward.
2) needless virtualized function call.
3) needless extra function call.
4) Feels very java instead of go.  The only reason we even have a TagList method and helpers is so we can accept different types for encoding. 
3) It keeps the types clean, instead of having it return an interface that then has a method to return a tag list, it is clear what it is returning.

clientoptions.go
docmerlin: Since each type of compression, has its own options, this would complicate stuff and make it harder to use imo.  You couldn't just have a WithCompression(compressionType, level int) because different compression formats use very different levels, it isn't always an integer 0-9, and sometimes they are unrelated to anything... for example if we end up supporting ZStd, we would need to pass in a dictionary etc.  Plus, I think the chance of us adding other compression schemes is pretty high... if just because gzip isn't great on embedded devices so something like lz4 would be useful.

Gzip would require 9 different constants to use a `WithCompression(compressionType)` function for example.

How about instead I just return an error with a helpful message  if someone tries to use incompatible options?

models.go
docmerlin: Yes, if they are manually making their own RowMetric.  I can think of lots of reasons to do this, instead of using the Add methods.

models.go
danielnelson: What's a virtualized function call in Go?

Returning a concrete type could be nice but I think then it should be a copy of the data.  If this type is meant to allow users to use their own unlying implementation, it seems that returning a concrete type forces a copy in most cases.  Maybe this should be `[]Tag`?

models.go
danielnelson: Why don't we just sort when we encode the data?

clientoptions.go
danielnelson: It's simple, just have options to your options. ;)  Just kidding of course, yeah I guess the error is best.

One thing I'm wondering now is why have WithNoCompression at all?  Isn't it the default?

docmerlin: The default is to use Gzip.  I was asked to make this the default, and it makes sense to me.  The only reason to not compress is if you are resource constrained on cpu.

models.go
docmerlin: A virtualized function call means it has to do extra work on the method call, and it normally can't be inlined.
Internally it stores the pointers so it is only a copy of the pointer values not of the value the pointer points to.
I'm not sure what making it a copy gives us, besides the extra copying.

PR #13, opened 2019-05-20
chore(readme): update readme to direct readers to 1.x client

### LINKED ISSUE #10: docs: Update README to clearly mention the 1.x client repository
```
package main

import (
	"context"
	"log"
	"net/http"
	"time"
)

func main() {

	var myHTTPClient *http.Client

	influx, err := influxdb.New(myHTTPClient, influxdb.WithAddress("http://localhost:8086"), influxdb.WithToken("mytoken"))
	if err != nil {
		panic(err) // error handling here, normally we wouldn't use fmt, but it works for the example
	}

	// we use client.NewRowMetric for the example because its easy, but if you need extra performance
	// it is fine to manually build the []client.Metric{}.
	myMetrics := []influxdb.Metric{
		influxdb.NewRowMetric(
			map[string]interface{}{"memory": 1000, "cpu": 0.93},
			"system-metrics",
			map[string]string{"hostname": "hal9000"},
			time.Date(2018, 3, 4, 5, 6, 7, 8, time.UTC)),
		influxdb.NewRowMetric(
			map[string]interface{}{"memory": 1000, "cpu": 0.93},
			"system-metrics",
			map[string]string{"hostname": "hal9000"},
			time.Date(2018, 3, 4, 5, 6, 7, 9, time.UTC)),
	}

	// The actual write..., this method can be called concurrently.
	if err := influx.Write(context.Background(), "my-awesome-bucket", "my-very-awesome-org", myMetrics...); err != nil {
		log.Fatal(err) // as above use your own error handling here.
	}
	influx.Close() // closes the client.  After this the client is useless.
}

```
ERROR `2019/05/10 01:07:57 json: cannot unmarshal number into Go value of type influxdb.genericRespError` exit status 1

AlirieGray:
Closes #10 

This PR updates the readme to clarify to the reader that this client only works for the 2.x line, and includes a link to the 1.x client library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # influxdb-client-go
-A home for InfluxDB’s 2.x's golang client. 
+
+A home for InfluxDB’s 2.x's golang client. This client is not compatible with InfluxDB 1.x--if you are looking for the 1.x golang client you can find it [here](https://github.com/influxdata/influxdb1-client).
 
 ## Pre-Alpha Warning!
 This is not even close to ready for prod use.


README.md
```diff
@@ -1,5 +1,6 @@
 # influxdb-client-go
-A home for InfluxDB’s 2.x's golang client. 
+
+A home for InfluxDB’s 2.x's golang client. This client is not compatible with InfluxDB 1.x--if you are looking for the 1.x golang client you can find it here: https://github.com/influxdata/influxdb1-client
```
docmerlin: make the url a link:
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links

PR #24, opened 2019-06-30
[F] fixbug write data fail when using line protocol

### LINKED ISSUE #23: how can i set precision when influx.Write()?

mouyong:
fix #23


mouyong:
move to #25

PR #26, opened 2019-06-30
[F] fixbug write data fail when using line protocol

mouyong:
api need the param

#23 

https://v2.docs.influxdata.com/v2.0/write-data/#write-data-using-the-influxdb-v2-api


docmerlin:
@mouyong  hey I don't think this PR does anything, mind if I close it?

mouyong:
@docmerlin ok.

PR #28, opened 2019-07-12
refactor: writers package (previously: Data Race On Stop)

GeorgeMac:
Closes: #39, #31 and #29

This PR now introduces a few fundimental reorganizations of resposibility. Prior to this change:
1. The `(*influxdb.Client).Write` handled its own retries + backoff logic
2. A `*influxdb.LPWriter` handled buffering and periodic flushing of metrics to a target bucket to `(*influxdb.Client).Write`.

This PR does the following:
1. remove retry + backoff functionality from `(*influxdb.Client).Write`.
2. Introduces some new types in a package `writer` which replace the `*LPWriter`.

The new types are as follows:

- `*BucketWriter` decorates a client, an organisation and a bucket and simplifies the `Write(org, bucket, ...metric)` to a simple `Write(...Metric)`.
- `*BufferedWriter` decorates the simpler Writer interface and buffers up writes. When buffer is exceeded it automatically flushes. Otherwise `writer.Flush()` can be called manually. 
- `*PointWriter` decorates a buffered writer, makes it safe for concurrent use and ensure Flush is called automatically and periodically. It ensures flush is called even if the buffer is not exceeded after an interval has ellapsed since the last call to flush. The last call to flush could have either been since last period or since buffer exceeding caused a flush.

The following is taken from the doc.go example:

```go
import (
	"github.com/influxdata/influxdb-client-go/writer"
)

func main() {
	cli, _ := influxdb.New("http://localhost:9999", "some-token")

	wr := writer.New(cli, influxdb.Organisation("influx"), influxdb.Bucket("default"), writer.WithBufferSize(10))

	wr.Write(influxdb.NewRowMetric(
		map[string]interface{}{
			"value": 16,
		},
		"temp_celsius",
		map[string]string{
			"room": "living_room",
		},
		time.Now(),
	),
		influxdb.NewRowMetric(
			map[string]interface{}{
				"value": 17,
			},
			"temp_celsius",
			map[string]string{
				"room": "living_room",
			},
			time.Now(),
		))

	wr.Close()
}

```

Will compose all the writers correctly and start the periodic flushing.

This is all demonstrated in the doc.go:

https://github.com/influxdata/influxdb-client-go/blob/73fd2441416a3216a09b1a1ad7aa0f8e1c17aa2b/writer/doc.go#L1-L59

TODO:

- [x] implement `*BucketWriter`
- [x] test `*BucketWriter`
- [x] implement `*BufferedWriter`
- [x] test `*BufferedWriter`
- [x] implement `*PointWriter`
- [x] test `*PointWriter`

---

Original Issue:

There is a data race on stop channel due to its complete replacement:

https://github.com/influxdata/influxdb-client-go/blob/de2584b400a81d0eaf88ad865d3fd4806accda2b/writer.go#L261-L264

Two things happen next when Stop is called.

1. `Write` doesn't obtain the lock before attempting to checked if `stop` is closed so we have a race on the reference here. Write could attempt to recv on stop while `Stop` is replacing with a new channel (I believe this is what the race detector is demonstrating).

2. The `Start` flush loop `Flush` may be busy flushing while `Stop` is replacing the `stop` channel and so it may be swapped back to a non-closed one on return:

https://github.com/influxdata/influxdb-client-go/blob/de2584b400a81d0eaf88ad865d3fd4806accda2b/writer.go#L117-L124

This means that the call to `Stop` may block forever as this flush loop may not exit and call Done on the waitgroup.

TODO:

- [x] enable race detector in CI
- [x] consider redesigning API to not allow Start/Stop but instead use Start/Close and make the buffered writer use once
- [ ] otherwise, rearchitect Stop. Ensure Stop follows Start follow Stop follows Start consistently.


writer/bucket.go
```diff
@@ -0,0 +1 @@
+package writer
```
GeorgeMac: note to self remove this

GeorgeMac:
@docmerlin my original intent was that it would enforce compile time ordering of org before bucket. I hate when you have a function and two arguments next to each other are the same type 😂 Just begging for someone to read the docs wrong and type them in the wrong order.

When I started this PR it was bucket then org and not for want of getting into a haystack / needle type debate, I figured it makes more sense from a heirarchy perspective that it goes org > bucket.
Partucularly for Write which takes org, bucket and then metric (or measurement). This way the type heirachy descends the same way our domain model does.

But then I realized if I leave this as strings then when you update to the new version of the client, bam, no compile errors and bucket and org have flipped. Breaking all your stuff.
So I wanted to enforce some compile time ordering.

All that said, you can pass these in as strings without the retype and Go will infer and it still compiles so this doesn't fix anything 😂

TL;DR

I swapped `org` and `bucket` argument order do we:
1. put them back in the original order
2. leave them in this new order, put them back to strings and not worry about breaking things
3. make org and bucket more complex type like a struct type with a Name field in order to create compile-time order concrete
4. do nothing

Also open to other ideas.

go.mod
```diff
@@ -4,4 +4,6 @@ require (
 	github.com/google/go-cmp v0.2.0 // test dependency
 	github.com/influxdata/flux v0.0.0-20190620184636-886e3c28388d // test dependency
 	github.com/influxdata/line-protocol v0.0.0-20190509173118-5712a8124a9a
+	github.com/stretchr/testify v1.3.0
```
docmerlin: add  // test dependency 
comment.  It helps when keeping track.

e2e_test.go
```diff
@@ -133,9 +134,11 @@ func TestE2E(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
 	b, err := ioutil.ReadAll(r)
 	if err != nil {
 		t.Fatal(err)
 	}
+
 	fmt.Println(string(b))
```
docmerlin: Is this a debug print?

GeorgeMac: I think this was there before. The e2e test seems to just do the query and print it out.
I just added the whitespace as I have a thing about gutters after close curly brace 😂

go.mod
GeorgeMac: nice! will do that!

e2e_test.go
GeorgeMac: I think a bigger piece needs to come next to do some real assertions on the returned data. I will leave this there until then.

PR #36, opened 2019-07-19
fix issue #35

### LINKED ISSUE #35: LPWriter.org is never assigned
When `org` is passed [here](https://github.com/influxdata/influxdb-client-go/blob/develop/writer.go#L65) it is not used inside the function. Then when writing the metrics `w.org` [is used](https://github.com/influxdata/influxdb-client-go/blob/develop/writer.go#L174), which is always empty.

docmerlin:
closes #35

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -67,7 +67,7 @@ func (c *Client) NewBufferingWriter(bucket string, org string, flushInterval tim
 	if onError == nil {
 		onError = func(_ error) {}
 	}
-	w := &LPWriter{c: c, buf: switchableBuffer{&bytes.Buffer{}}, flushSize: flushSize, flushInterval: flushInterval, stop: make(chan struct{}), onError: onError}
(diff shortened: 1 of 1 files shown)


writer.go
```diff
@@ -67,7 +67,7 @@ func (c *Client) NewBufferingWriter(bucket string, org string, flushInterval tim
 	if onError == nil {
 		onError = func(_ error) {}
 	}
-	w := &LPWriter{c: c, buf: switchableBuffer{&bytes.Buffer{}}, flushSize: flushSize, flushInterval: flushInterval, stop: make(chan struct{}), onError: onError}
+	w := &LPWriter{c: c, buf: switchableBuffer{&bytes.Buffer{}}, flushSize: flushSize, flushInterval: flushInterval, stop: make(chan struct{}), onError: onError, bucket: bucket,org: org}
```
GeorgeMac: I think you need a `gofmt` here. Perhaps worth adding that to the circleci check as well.

PR #38, opened 2019-07-23
doc: Fix code block for QueryCSVResult.Next()

benbjohnson:
Indents the code for the godoc so that it displays properly.


benbjohnson:
Screen Shot 2019-07-23 at 8 37 45 AM

PR #50, opened 2019-09-09
writer.RetryWriter implementation

### LINKED ISSUE #46: Add Retry Implementation of writer.BucketMetricWriter
The root `influxdb.Client` implements the `writer.BucketMetricWriter` interface (https://github.com/influxdata/influxdb-client-go/blob/develop/writer/writer.go#L11).
A new decorating implementation is required which handles automatically retrying on error conditions.

This was previously implemented in the `influxdb.Client`, however, it was removed in the buffered writer refactor. This should now be reinstated as a decorator for the `writer.BucketMetricWriter` interface.

GeorgeMac:
closes #46

This reinstates the ability to automatically retry writes based on known transient error conditions.

See updates to docs.go:
```go
import (
	"time"

	"github.com/influxdata/influxdb-client-go/writer"
)

func main() {
	var (
		cli, _ = influxdb.New("http://localhost:9999", "some-token")
		bucket = "default"
		org    = "influx"
	)

	// construct a writer with 3 maximum attempts per call to Write and linear backoff derived from number of attempts
	// i.e. a first attempt is followed by a 1 second delay before the second attempt
	// a second attempt is followed by a 2 second delay before the third attempt
	var (
		retryOpts = []writer.RetryOption{writer.WithMaxAttempts(3), writer.WithBackoff(writer.LinearBackoff(time.Second))}
		wr        = writer.New(cli, bucket, org, writer.WithRetries(retryOpts...))
	)
}
```

Namely:
- unavailable
- too many requests (rate limiting)

TODO:

- [x] implement basic retry behavior with max attempts
- [x] implement backoff which respects Retry-After response header
- [x] implement backoff which is configured via a backoff function
- [x] configure and wire retry writer into `writer.New()` response type
- [x] more tests around `write.go` handling of error parsing

### DIFF
--- a/writer/retry.go
+++ b/writer/retry.go
@@ -63,6 +63,7 @@ func (r *RetryWriter) Write(m ...influxdb.Metric) (n int, err error) {
 				// given retry-after is configured attempt to sleep
 				// for retry-after seconds
 				r.sleep(time.Duration(*ierr.RetryAfter) * time.Second)
+				continue
 			}
 
 			// given a backoff duration > 0
--- a/writer/retry.go
+++ b/writer/retry.go
@@ -8,13 +8,19 @@ import (
 
 const defaultMaxAttempts = 5
 
+// BackoffFunc is a function which when called with an
+// attempt number returns a duration which should be
+// waited for until a subsequent attempt is made
+type BackoffFunc func(attempt int) time.Duration
+
 // RetryWriter is a metrics writers which decorates other
 // metrics writer implementations and automatically retries
 // attempts to write metrics under certain error conditions
 type RetryWriter struct {
 	MetricsWriter
 
-	sleep func(time.Duration)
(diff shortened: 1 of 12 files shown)


writer/retry.go
```diff
@@ -0,0 +1,107 @@
...
+				// given retry-after is configured attempt to sleep
+				// for retry-after seconds
+				r.sleep(time.Duration(*ierr.RetryAfter) * time.Second)
+			}
+
+			// given a backoff duration > 0
+			if duration := r.backoff(i + 1); duration > 0 {
+				// call sleep with backoff duration
+				r.sleep(duration)
+			}
```
GeorgeMac: note to self: I need to add a test which captures the fact that we should only sleep once here. Then add the fix to make it only sleep once per error.

PR #56, opened 2019-10-15
Fix query: transport when gzip

cnbailian:
When the bytes of the response are larger than a certain size, the api will respond to `Content-Encoding: gzip`, so `gzip.NewReader` should be used when` Accept-Encoding` is `gzip`.

```
However, if the user explicitly requested gzip it is not automatically uncompressed.
```

### DIFF
--- a/query.go
+++ b/query.go
@@ -2,6 +2,7 @@ package influxdb
 
 import (
 	"bytes"
+	"compress/gzip"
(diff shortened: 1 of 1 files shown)


docmerlin:
@cnbailian thanks man, we are going to add some tests around that, but thanks for fixing it.

PR #58, opened 2019-10-18
Release v0.1.2

GeorgeMac:
Cut a new v0.1.2 release containing the fix to bucket and org argument order.


GeorgeMac:
👍 There is one. Thinking ahead we should integrate the `writers` package into the end to end tests. That would have also caused this. I will create a card for that.

PR #61, opened 2019-10-21
Add WithContext config option to writers package

GeorgeMac:
This allows the threading of a `context.Context` to be made for all batches written to a `writer.PointWriter` constructed with `New(...)`

Each time `BufferedWriter.Flush` is called this `context.Context` will be used in the call to `client.Write(ctx, ...)`.

### DIFF
--- a/writer/options.go
+++ b/writer/options.go
@@ -1,9 +1,13 @@
 package writer
 
-import "time"
+import (
+	"context"
+	"time"
+)
 
 // Config is a structure used to configure a point writer
 type Config struct {
+	ctxt          context.Context
 	size          int
 	flushInterval time.Duration
 	retry         bool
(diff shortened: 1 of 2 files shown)


gianarb:
I am not sure if I like this idea. My expectation is that everytime a function needs a context it gets its own context. Not sure why it should inherit one from a configuration

GeorgeMac:
Good question @gianarb 
The reason is because a call to `Write(m)` on this type doesn't mean that the `Write` call goes directly to the underlying client. It is instead buffered and gets flushed in a bigger batch. This means we could potentially have > 1 contexts per call to `client.Write` which would need some kind of reconciliation or merging of contexts which is nasty.

The simplest approach here is to use one context per `writers.PointWriter` (which `New()` returns). Making the whole buffered writers life-time batched.

gianarb:
Ah got it @GeorgeMac ! Thanks for the explanation! I think at this point we can have a context at the client level

PR #70, opened 2019-12-11
Fix/68

### LINKED ISSUE #68: client.QueryCSV fails for multiple yields
I have a flux query which returns 2 yields in a single script as follows:
(Assuming all the values are sent)
```flux
workbench = from(bucket: "test")
  |> range(start: v.timeStart, stop: v.timeEnd)
  |> filter(fn: (r) => r._measurement == "mem")
  |> filter(fn: (r) => r.host == v.host)

count = workbench
    |> count()
    |> yield(name:"count")
data = workbench
    |> limit(n: v.limit, offset : v.offset)
    |> sort(columns: ["_time"], desc: v.isDescending)
    |> yield(name:"data")
```

From the resulting query I get 2 series in the HTTP request. But the package fails to 
read the second series returning the error `record on line 61: wrong number of fields`. So I tried setting the csvreader with `FieldsPerRecord` as -1 (as per [this](https://golang.org/pkg/encoding/csv/#Reader)). It fails for that too. Need help.

docmerlin:
fixes #68 

This should allow for unmarshaling when you have tables with different dimensions in the query table stream.

### DIFF
--- a/query_test.go
+++ b/query_test.go
@@ -7,11 +7,11 @@ import (
 	"testing"
 	"time"
 
-	cmp "github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp"
 )
 
 func mustParseTime(s string) time.Time {
-	t, err := time.Parse(time.RFC3339Nano, s)
+	t, err := time.Parse(time.RFC3339, s)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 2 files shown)


query_test.go
```diff
@@ -7,11 +7,11 @@ import (
 	"testing"
 	"time"
 
-	cmp "github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp"
 )
 
 func mustParseTime(s string) time.Time {
-	t, err := time.Parse(time.RFC3339Nano, s)
+	t, err := time.Parse(time.RFC3339, s)
```
AlirieGray: is it ok that we're losing nanosecond precision here? I assume so but wanted to check

docmerlin: Yah, it works fine for the tests.

Issue #72, opened 2020-01-09
feature: handle returned http.StatusRequestEntityTooLarge, in the various automated buffered writers that auto send.


docmerlin:
handle returned http.StatusRequestEntityTooLarge, in the various automated buffered writers that auto send.

docmerlin:
This would target the writers in
 https://github.com/influxdata/influxdb-client-go/tree/develop/writer/

PR #77, opened 2020-02-07
feat: Retry writer should split too-big payloads

docmerlin:
This changes the retry writer to split excessively large payloads when
the influxdb returns an http.StatusRequestEntityTooLarge response.

Closes: https://github.com/influxdata/influxdb-client-go/issues/72

### DIFF
--- a/e2e_test.go
+++ b/e2e_test.go
@@ -8,7 +8,7 @@ import (
 	"testing"
 	"time"
 
-	influxdb "github.com/influxdata/influxdb-client-go"
+	"github.com/influxdata/influxdb-client-go"
 )
 
 var e2e bool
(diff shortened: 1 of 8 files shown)


e2e_test.go
```diff
@@ -17,7 +18,11 @@ var fuzz bool
 func init() {
 	flag.BoolVar(&e2e, "e2e", false, "run the end tests (requires a working influxdb instance on 127.0.0.1)")
 	flag.BoolVar(&fuzz, "fuzz", false, "(Not Implemented) run the end tests (requires a working influxdb instance on 127.0.0.1)")
+}
+
+func TestMain(m *testing.M) {
 	flag.Parse()
+	os.Exit(m.Run())
```
GeorgeMac: Take it or leave it: `testing` package calls `flag.Parse()` automatically. So entire test main here can be dropped.

errors.go
```diff
@@ -18,6 +18,8 @@ const (
 	ETooManyRequests     = "too many requests"
 	EUnauthorized        = "unauthorized"
 	EMethodNotAllowed    = "method not allowed"
+	ETooLarge            = "request too large"
+	StatusETooLarge      = 413
```
GeorgeMac: Is this needed?

PR #79, opened 2020-02-28
extract WriteBuffer from Write to allow writing raw line protocol

lachie:
I have raw line protocol that I'd like to forward through to influx via
the client.

I don't want to parse it into `Metric`'s only to encode it back to bytes.

I could implement my own client in raw net/http, but I'd like to be able
to use some of the official client's config & careful connection
handling.


vlastahajek:
We released a new client with the updated API.  Writing a line protocol record  is possible via
WriteRecord method. Check examples in Readme, please.

PR #86, opened 2020-03-24
Expose CSV column properties

ryantxu:
I am working on a grafana plugin that will read CSV results from influx and convert the results to an arrow table.

The QueryCSVResult is almost perfect... except that I need access to the private properties.  This PR:
* makes DataType, Group, DefaultVals public
* adds a helper function to create QueryCSVResult from a byte buffer.  This is helpful for testing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,5 @@
 module github.com/influxdata/influxdb-client-go
 
-go 1.14
-
 require (
 	github.com/google/go-cmp v0.2.0 // test dependency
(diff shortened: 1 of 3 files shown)


vlastahajek:
We released a new client with the updated API.  QueryCSVResult  was replaced with [QueryTableResult](https://github.com/influxdata/influxdb-client-go/blob/master/query.go#L162)  with hi-level API and access to all properties

ryantxu:
thank you!

PR #136, opened 2020-06-19
feat: adding possibility to set default tags

### LINKED ISSUE #135: Default Tags
Add possibility to set default tags

vlastahajek:
Closes #135

Added possibility set default tags

- [X] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/api/write/options.go
+++ b/api/write/options.go
@@ -25,6 +25,8 @@ type Options struct {
 	precision time.Duration
 	// Whether to use GZip compression in requests. Default false
 	useGZip bool
+	// Tags added to each point during writing. If a point already has a tag with the same key, it is left unchanged.
+	defaultTags map[string]string
 }
 
 // BatchSize returns size of batch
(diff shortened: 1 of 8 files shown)


api/write/options.go
```diff
@@ -25,6 +25,8 @@ type Options struct {
 	precision time.Duration
 	// Whether to use GZip compression in requests. Default false
 	useGZip bool
+	// Point settings, default tags
+	pointSettings *PointSettings
```
sranka, line 29: Wouldn't be better to simply use `defaultTags` herein. Precision (as a point setting) is also part of `Options`.

internal/write/service.go
```diff
@@ -150,14 +151,67 @@ func (w *Service) WriteBatch(ctx context.Context, batch *Batch) error {
...
 	var buffer bytes.Buffer
 	e := lp.NewEncoder(&buffer)
 	e.SetFieldTypeSupport(lp.UintSupport)
 	e.FailOnFieldErr(true)
 	e.SetPrecision(w.writeOptions.Precision())
+	var m lp.Metric
 	for _, point := range points {
-		_, err := e.Encode(point)
+		if len(w.writeOptions.PointSettings().DefaultTags()) > 0 {
+			m = &pointWithDefaultTags{
```
sranka, line 207: it might be better to reuse `pointToEncode` outside the for loop

internal/write/service_test.go
```diff
@@ -0,0 +1,52 @@
...
+	p.AddField("i", 1)
+
+	s, err = srv.EncodePoints(p)
+	require.Nil(t, err)
+	assert.Equal(t, "x,dt1=val1,xt=1,zdt=val2 i=1i\n", s)
+	assert.Len(t, p.TagList(), 1)
+
+	p = write.NewPointWithMeasurement("d")
+	p.AddTag("id", "1")
+	// do not overwrite point tag
```
sranka, line 43: well done

options.go
```diff
@@ -148,6 +148,14 @@ func (o *Options) HttpOptions() *http.Options {
 	return o.httpOptions
 }
 
+// AddDefaultTag adds a default tag. DefaultTags are added to each written point.
+// If a tag with the same key already exist it is overwritten.
+// If a point already defines a such tag, it is left unchanged
```
sranka, line 153: defines such a tag

api/write/options.go
vlastahajek: I'm 100% for this. This was my first idea as well, but when I looked at Java and Js client I see default tags are part of PointSetting

sranka: PointSettngs serve different purpose in JS, it is not an API configuration iface.

PR #138, opened 2020-06-24
fix: errors from InfluxDB 1.8 being empty

diamondburned:
This PR fixes errors from InfluxDB 1.8 being empty. It does this by falling back to reading the `X-Influxdb-Error`, and if that doesn't work, then `http.Error` would return the unexpected status code inside the error message.

Before PR:

before PR screenshot

After PR:

after PR screenshot

- [x] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x] Tests pass


diamondburned:
I genuinely have no idea why the tests are failing somewhere else after only adding 2 more tests.

vlastahajek:
Thanks for the update. Don't worry about the tests. There is a change between beta 12 and beta 13 that made some tests failing.

PR #144, opened 2020-07-04
Feature: Add log api

### LINKED ISSUE #104: Make logger configurable
Usage of `log.Logger` is private and hard-wired. Would be helpful to set a custom logger.

andig:
Fix #104:

- expose public `Logger` api
- implement a default console logger
- add an internal `LogAdapter` that connected client with public logger
- make `SetDebugLevel` part of `LogAdapter`

- [x] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x] Tests pass


vlastahajek:
@andig, thanks for the PR! 
The current logging mechanism and API  will be examined, so stay tuned.

andig:
> these are tightly connected to each other

- `api/log` currently (and historically) contains the internal-only logger implementation- I'd like to move that to `/internal/log`
- package needs to include `/internal/log` to use `log.Log`. It is easy to mistake this with `api/log`- let's move `api/log` to `api`, containing only the interface (and the instance)
- last we might consider if we should move the `Log` instance to a client option. This would be a bit more invasive.

Makes sense?

vlastahajek:
I plan to use Uber zap.Logger as a default Logger. As it has not its own interface, it still makes sense to add a logger interface. The best place is to put it to the `/log` folder, as Pavel suggested. `/api` folder is intended for the server REST API interfaces and implementations. 
I still don't have a clear idea of how it should finally look like, I will get to this at the beginning of next week.  
There will be regularly scheduled release with mostly bug fixes tomorrow, so the logging improvement will be, for sure now, in the release after.

internal/log/logger.go
```diff
@@ -5,61 +5,68 @@
...
+	}
 }
 
-func (l *Logger) Error(msg string) {
-	log.Print("[E]! ", msg)
+func (l *LogAdapter) Error(msg string) {
+	if log.Log != nil {
+		log.Log.Error(msg)
+	}
 }
```
vlastahajek, line 72: There should be the default implementation with the original code.  `init()` could initially set `logger.Log`

andig:
@vlastahajek appreciate your comments but not sure I fully follow. I was trying to achieve 3 goals:

- have a public logger interface
- separate setting the debug level from the logger interface (it should really be part of client options, but thats another topic)
- keep the default implementation (using console logger)
- clean the implementation's interface such that the [E] prefixes become part of the default implementation, not of the logger interface

If that makes sense I could take another stab at it

vlastahajek:
@andig, my comments mainly target the fact your proposal introduces two logging objects, the default implementation, `logger`, and `LogAdapter`. 

In fact, in the actual state, the `internal/log/Logger` is already a log adapter, because it uses the standard `log` package under the hood. If you need to override the log target, which is essentially what `Logger` interface in your design means, you could just replace the actual log writer in the standard `log` package, without the necessity to have a logger interface.

As the filtering capability, and also prefixing, is part of the logger responsibilities, the logger should handle that directly.
`SetLogLevel` of Client options will simply forward this to the current logger.

andig:
> As the filtering capability, and also prefixing, is part of the logger responsibilities, the logger should handle that directly.
SetLogLevel of Client options will simply forward this to the current logger.

Prefixing: ack

Filtering: I think it comes down to where you want the SetLogLevel handled. Imho that is part of the client duties- it should only log what it is supposed to log. With that approach the adapter is needed or a lot if if/elses across the codebase. Same if you want to be able to nil the logger. I'd propose to keep the log interface slim and handle SetLogLevel in this library.

vlastahajek:
As Logger defines what levels are supported, Logger should be responsible for filtering of those levels.
This is how third-party loggers behave.

vlastahajek:
@andig, thanks for your effort. I had to complete this, so I implemented that according to my thoughts.

andig:
Thank you- just didn‘t get around to actually finish this, sorry.

PR #165, opened 2020-07-28
feat: allow overriding the http.Client for the http service

jsternberg:
This allows overriding the http.Client for the http service. If the
client is overridden, the other options are ignored. This allows further
customization such as modifying the underlying http.Transport or the
`CheckRedirect` function.

### DIFF
--- a/api/http/options.go
+++ b/api/http/options.go
@@ -7,14 +7,49 @@ package http
 
 import (
 	"crypto/tls"
+	"net"
+	"net/http"
+	"time"
 )
 
 // Options holds http configuration properties for communicating with InfluxDB server
 type Options struct {
+	// HTTP client. Default is http.DefaultClient.
+	httpClient *http.Client
 	// TLS configuration for secure connection. Default nil
 	tlsConfig *tls.Config
 	// HTTP request timeout in sec. Default 20
 	httpRequestTimeout uint
+}
+
+// HTTPClient returns the http.Client that is configured to be used
+// for HTTP requests. It will return the one that has been set using
+// SetHTTPClient or it will construct a default client using the
+// other configured options.
+func (o *Options) HTTPClient() *http.Client {
+	if o.httpClient == nil {
+		o.httpClient = &http.Client{
+			Timeout: time.Second * time.Duration(o.HTTPRequestTimeout()),
+			Transport: &http.Transport{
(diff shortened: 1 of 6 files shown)


api/http/options.go
```diff
@@ -7,16 +7,51 @@ package http
...
+	return &http.Client{
+		Timeout: time.Second * time.Duration(o.HTTPRequestTimeout()),
+		Transport: &http.Transport{
+			DialContext: (&net.Dialer{
+				Timeout: 5 * time.Second,
+			}).DialContext,
+			TLSHandshakeTimeout: 5 * time.Second,
+			TLSClientConfig:     o.TLSConfig(),
+		},
+	}
```
vlastahajek, line 42: Why not setting newly created HTTP client to `o.httpClient`? It would allow returning the HTTP client used by the library.

jsternberg: I can change it to do that.

vlastahajek:
Thanks for the PR! Update also changelog, please.

options.go
```diff
@@ -110,6 +112,24 @@ func (o *Options) SetUseGZip(useGZip bool) *Options {
 	return o
 }
 
+// HTTPClient returns the http.Client that is configured to be used
+// for HTTP requests. It will return the one that has been set using
+// SetHTTPClient or it will construct a default client using the
+// other configured options.
+func (o *Options) HTTPClient() *nethttp.Client {
+	return o.httpOptions.HTTPClient()
+}
```
vlastahajek, line 121: This should be tested in the `/options_test.go`

options.go
```diff
@@ -110,6 +112,24 @@ func (o *Options) SetUseGZip(useGZip bool) *Options {
...
+// SetHTTPClient will configure the http.Client that is used
+// for HTTP requests. If set to nil, an HTTPClient will be
+// generated.
+//
+// Setting the HTTPClient will cause the other HTTP options
+// to be ignored.
+func (o *Options) SetHTTPClient(c *nethttp.Client) {
+	o.httpOptions.SetHTTPClient(c)
+}
+
```
vlastahajek, line 132: This should be also tested in the `/options_test.go`

options.go
jsternberg: Done.

PR #184, opened 2020-08-14
fix: Closing body at the end of the result (#183)

### LINKED ISSUE #183: Client is leaking TCP connections
I discovered this issue with the grafana flux datasource, but traced it back to the client. It seems TCP connections are not closed after calling `client.Close()`.

Output of the program below:
```
...
Query returned 181 rows
Open TCP connections:       59
Query returned 181 rows
Open TCP connections:       60
Query returned 181 rows
Open TCP connections:       61
Query returned 181 rows
Open TCP connections:       62
...
```

Please run this go program to see for yourself (when on linux, change .8086 to :8086 for the netstat command).
```go
package main

import (
	"context"
	"fmt"
	"log"
	"os/exec"

	influxdb2 "github.com/influxdata/influxdb-client-go"
)

func main() {
	for {
		client := influxdb2.NewClient("https://demo.factry.io:8086", "flux:xulf")
		queryAPI := client.QueryAPI("factry")

		query := `from(bucket: "simulation_tanksystem")
							|> range(start: -2h, stop:-30m)
							|> filter(fn: (r) => r._measurement == "WP1_fill_level")
							|> aggregateWindow(every: 30s, timeSrc: "_start", fn: mean, createEmpty:false)`

		// get parser flux query result
		result, err := queryAPI.Query(context.Background(), query)
		if err != nil {
			fmt.Println(err)
			return
		}
	        defer result.Close()

		count := 0
		// Use Next() to iterate over query result lines
		for result.Next() {
			// Observe when there is new grouping key producing new table
			count++
		}
		fmt.Printf("Query returned %v rows\n", count)
		if result.Err() != nil {
			fmt.Printf("Query error: %s\n", result.Err().Error())
		}

		client.Close()

		getOpenConns()
	}
}

func getOpenConns() {
	out, err := exec.Command("sh", "-c", `netstat -atn | grep ".8086" | wc -l`).Output()
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("Open TCP connections: %s", out)
}

```

vlastahajek:
Closes #183

Closing result (body) at the end of the result

## Checklist

- [X] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/api/query.go
+++ b/api/query.go
@@ -220,6 +220,8 @@ const (
 // During the first time it is called, Next creates also table metadata
 // Actual parsed row is available through Record() function
 // Returns false in case of end or an error, otherwise true
+// Close() is called automatically at the end of the result or in case of an error.
+// If result is not read to the end, it must be closed explicitly.
 func (q *QueryTableResult) Next() bool {
 	var row []string
 	// set closing query in case of preliminary return
@@ -241,7 +243,7 @@ func (q *QueryTableResult) Next() bool {
 readRow:
 	row, q.err = q.csvReader.Read()
 	if q.err == io.EOF {
-		q.err = nil
+		q.err = q.Close()
 		return false
 	}
 	if q.err != nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Bug fixes 
 1. [#175](https://github.com/influxdata/influxdb-client-go/pull/175) Fixed WriteAPIs management. Keeping single instance for each org and bucket pair.
+1. [#184](https://github.com/influxdata/influxdb-client-go/pull/184) Closing result at end of reading to avoid leaked connections.
 
 ## 1.4.0 [2020-07-17]
 ### Breaking changes


api/query.go
```diff
@@ -241,7 +243,7 @@ func (q *QueryTableResult) Next() bool {
 readRow:
 	row, q.err = q.csvReader.Read()
 	if q.err == io.EOF {
-		q.err = nil
+		q.err = q.Close()
```
aknuds1, line 246: Actually, I have a question here. Shouldn't `q.Close()` be called also on line 250?

vlastahajek:
I actually got confused. This fix shouldn't be necessary and the problem is probably somewhere else. Looking again at the code, there is the deferred closer which is called in all cases if not returning true, it should be called at the end.

aknuds1:
> I actually got confused. This fix shouldn't be necessary and the problem is probably somewhere else. Looking again at the code, there is the deferred closer which is called in all cases if not returning true, so it should be called at the end.

Ah, confusing indeed.

PR #188, opened 2020-08-15
Add clarification for influx 1.8

### LINKED ISSUE #182: Client does not work with compatibility endpoints
Using the then endpoint URL /api/v2 as described in the documentation does not work:
```
package main

import (
	"context"
	"fmt"

	influxdb2 "github.com/influxdata/influxdb-client-go"
)

func main() {

	client := influxdb2.NewClient("https://demo.factry.io:8086/api/v2", "token_here")
	queryAPI := client.QueryAPI("factry")

	query := `from(bucket: "simulation_tanksystem")
							|> range(start: -2h, stop:-30m)
							|> filter(fn: (r) => r._measurement == "WP1_fill_level")
							|> aggregateWindow(every: 30s, timeSrc: "_start", fn: mean, createEmpty:false)`

	// get parser flux query result
	_, err := queryAPI.Query(context.Background(), query)
	if err != nil {
		fmt.Println(err)
	}

}
```
this returns 404 not found. No using /api/v2, so just `influxdb2.NewClient("https://demo.factry.io:8086", "token_here")`, executes the query successfully. This behaviour is inconsistent with what both the influxdb 1.8 docs and this client's docs mention.

coussej:
Closes #182

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,7 @@ func main() {
   [InfluxDB 1.8.0 introduced forward compatibility APIs](https://docs.influxdata.com/influxdb/latest/tools/api/#influxdb-2-0-api-compatibility-endpoints) for InfluxDB 2.0. This allow you to easily move from InfluxDB 1.x to InfluxDB 2.0 Cloud or open source.
   
   Client API usage differences summary:
+ 1. Use the base URL when creating a new client, not the compatibility endpoint, so `host:port` instead of `host:port/api/v2`. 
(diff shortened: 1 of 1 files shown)


vlastahajek:
@coussej, Thanks for this PR. However, this is not the best place where to put it. This paragraph only summarizes key points when using this client for InfluxDB 1.8.

Using the base URL for serverURL is required every time. 

I've added #189 that mentions this on all main places

PR #210, opened 2020-11-09
fix: default http.client may cause many time_wait sockets

Chaiyq:


api/http/options.go
```diff
@@ -37,6 +37,8 @@ func (o *Options) HTTPClient() *http.Client {
 				}).DialContext,
 				TLSHandshakeTimeout: 5 * time.Second,
 				TLSClientConfig:     o.TLSConfig(),
+				MaxIdleConns:			100,
+				MaxIdleConnsPerHost:	100,
```
sranka, line 41: Why this? I guess that this is should be left unset.

vlastahajek, line 41:
```suggestion
				MaxIdleConnsPerHost:	100,
				IdleConnTimeout:            90 * time.Second,
```

sranka (approved):

PR #214, opened 2020-12-09
fix: write to errCh when point encoding fails

pbivrell:
Closes #

## Proposed Changes

Write the point encoding err to the errCh when the errCh is not nil.

## Checklist

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/api/write.go
+++ b/api/write.go
@@ -206,6 +206,9 @@ func (w *WriteAPIImpl) WritePoint(point *write.Point) {
 	line, err := w.service.EncodePoints(point)
 	if err != nil {
 		log.Errorf("point encoding error: %s\n", err.Error())
+		if w.errCh != nil {
+			w.errCh <- err
+		}
 	} else {
 		w.bufferCh <- line
 	}


api/write.go
```diff
@@ -206,9 +206,11 @@ func (w *WriteAPIImpl) WritePoint(point *write.Point) {
 	line, err := w.service.EncodePoints(point)
 	if err != nil {
 		log.Errorf("point encoding error: %s\n", err.Error())
-	} else {
-		w.bufferCh <- line
+		if w.errCh != nil {
+			w.errCh <- err	
+		}
 	}
+	w.bufferCh <- line
```
vlastahajek, line 213: `line` is empty in case of an error. It should be on the other branch of if

vlastahajek (changes requested):
Thanks for the PR. Just a little remark.

api/write.go
pbivrell: Ah good catch. Updated. Thanks.

vlastahajek (approved):

PR #218, opened 2020-12-15
docs: remove link to QueryTableResult

soolaugust:
As future change, remove link of QueryTableResult as it's unnecessary

## Proposed Changes

current link of QueryTableResult is wrong. should fix it.


vlastahajek:
@soolaugust, thanks for the PR! Please, rebase your branch and fix the commit message.

sranka:
This PR references a specific commit, it is stable but it can get old without a notice.  I think that it would better to simply use `QueryTableResult`. Rationalle: A specific link to the structure does not explain how to use it, a link to documentation or an example can.

soolaugust:
@sranka I agree with you. actually the following codes has explained how to use `QueryTableResult`. maybe the link is unnecessary. If developers want to know the detail, they will check the codes.

README.md
```diff
@@ -300,7 +300,7 @@ func main() {
 ```
 
 ### Queries
-Query client offers two ways of retrieving query results, parsed representation in [QueryTableResult](https://github.com/influxdata/influxdb-client-go/blob/master/query.go#L162) and a raw result string. 
+Query client offers two ways of retrieving query results, parsed representation in QueryTableResult and a raw result string.
```
sranka, line 303:
```suggestion
Query client offers retrieving of query results to a parsed representation in a `QueryTableResult` or to a raw result string. 
```

sranka (approved):

PR #227, opened 2021-01-22
feat(v3): Params, Client and New

vlastahajek:
## Proposed Changes

Adding initial implementation for Params and Client with New.

Package docs is still subject to further updates

## Checklist

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/inluxclient/client.go
+++ b/inluxclient/client.go
@@ -1,0 +1,73 @@
+// Copyright 2021 InfluxData, Inc. All rights reserved.
+// Use of this source code is governed by MIT
+// license that can be found in the LICENSE file.
+
+// Package influxclient provides client for InfluxDB server.
+package influxclient
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// Params holds the parameters for creating a new client.
+// The only mandatory field is ServerURL. AuthToken is also important
+// if authentication was not done outside this client.
+type Params struct {
+	// ServerURL holds the URL of the InfluxDB server to connect to.
+	// This must be non-empty. E.g. http://localhost:8086
+	ServerURL string
+
+	// AuthToken holds the authorization token for the API.
+	// This can be obtained through the GUI web browser interface.
+	AuthToken string
+
+	// HTTPClient is used to make API requests.
+	//
+	// This can be used to specify a custom TLS configuration
+	// (TLSClientConfig), a custom request timeout (Timeout),
+	// or other customization as required.
+	//
+	// It HTTPClient is nil, http.DefaultClient will be used.
+	HTTPClient *http.Client
+}
+
+// Client implements an InfluxDB client.
+type Client struct {
+	// Configuration params.
+	params Params
+	// Pre-created Authorization HTTP header value.
+	authorization string
+	// Cached base server API URL.
+	apiURL *url.URL
+}
+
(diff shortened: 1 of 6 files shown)


inluxclient/client.go
```diff
@@ -0,0 +1,93 @@
...
+type Params struct {
+	// ServerURL holds the URL of the InfluxDB server to connect to.
+	// This must be non-empty. E.g. http://localhost:8086
+	ServerURL string
+
+	// AuthToken holds the authorization token for the API.
+	// This can be obtained through the GUI web browser interface.
+	AuthToken string
+
+	// DefaultTags specifies a set of tags that will be added to each written
```
rogpeppe, line 34: How about leaving the parameters that are strictly writer-related out for now, and adding them when the relevant functionality is implemented? That way we can be more sure that we're not leaving around API that never gets hooked up. I'd include both `BatchSize` and `FlushInterval` in this suggestion.

inluxclient/client.go
```diff
@@ -0,0 +1,93 @@
...
+		c.params.HTTPClient = http.DefaultClient
+	}
+
+	serverAddress := params.ServerURL
+	if !strings.HasSuffix(serverAddress, "/") {
+		// For subsequent path parts concatenation, url has to end with '/'
+		serverAddress = params.ServerURL + "/"
+	}
+	// Prepare server API URL
+	serverURL, err := url.Parse(serverAddress)
```
rogpeppe, line 87: or just `url.Parse(serverAddress + "api/v2/")` ?

inluxclient/client.go
```diff
@@ -0,0 +1,93 @@
...
+	authorization string
+	// Cached base server API URL.
+	apiURL *url.URL
+}
+
+// New creates new Client with given Params, where ServerURL and AuthToken are mandatory.
+func New(params Params) (*Client, error) {
+	c := &Client{params: params}
+	if params.ServerURL == "" {
+		return nil, ErrEmptyServerURL
```
rogpeppe, line 72: I wouldn't bother exporting a specific error value for this (and making the API bigger as a result) - I'd just return `fmt.Errorf("empty influxclient ServerURL parameter")`

inluxclient/client.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Client implements an InfluxDB client.
+type Client struct {
+	// Configuration params.
+	params Params
+	// Pre-created Authorization HTTP header value.
+	authorization string
+	// Cached base server API URL.
+	apiURL *url.URL
```
rogpeppe, line 65: Given that you'll be needing to parse the URL again when making the request, I wonder if it might be best just to keep the URL in string form.

rogpeppe (approved):
LGTM modulo a few minor comments and suggestions, thanks!

inluxclient/client.go
vlastahajek: Ok

inluxclient/client.go
vlastahajek: Ok

inluxclient/client.go
vlastahajek: Ok

inluxclient/client.go
vlastahajek: The idea here is to avoid repeatedly parsing the complete URL string to the URL object, which will be used in specific API calls to construct endpoint URL with query params.

PR #230, opened 2021-02-05
feat(v3): Query part I. - QueryResults 

vlastahajek:
## Proposed Changes

First PR for Query: Adding `QueryResults` object, responsible for parsing Flux CSV stream and providing access to table columns and values.

## Checklist

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Influxdata, Inc.
+Copyright (c) 2021 Influxdata, Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/TODO.md
+++ b/TODO.md
@@ -399,37 +399,6 @@ type Filter struct {
 	AfterTime time.Time
 }
 
-type QueryResult struct {
-	// unexported fields.
-}
-
-// NextTable advances to the next table in the result.
-// Any remaining data in the current table is discarded.
-//
-// When there are no more tables, it returns false.
-func (r *QueryResult) NextTable() bool
-
-// NextRow advances to the next row in the current table.
-// When there are no more rows in the current table, it
-// returns false.
-func (r *QueryResult) NextRow() bool
-
-// Columns returns information on the columns in the current
-// table. It returns nil if there is no current table (for example
-// before NextTable has been called, or after NextTable returns false).
-func (r *QueryResult) Columns() []TableColumn
-
-// Err returns any error encountered. This should be called after NextTable
-// returns false to check that all the results were correctly received.
-func (r *QueryResult) Err() error
-
-// Values returns the values in the current row.
-// It returns nil if there is no current row.
-// All rows in a table have the same number of values.
-// The caller should not use the slice after NextRow
-// has been called again, because it's re-used.
-func (r *QueryResult) Values() []interface{}
-
 // Decode decodes the current row into x, which should be
 // a pointer to a struct. Columns in the row are decoded into
 // appropriate fields in the struct, using the tag conventions
--- a/annotatedcsv/reader.go
+++ b/annotatedcsv/reader.go
@@ -1,0 +1,340 @@
+// Copyright 2021 InfluxData, Inc. All rights reserved.
+// Use of this source code is governed by MIT
+// license that can be found in the LICENSE file.
+
+// Package annotatedcsv provides a reader for annotated CSV sources.
+// Annotated CSV must contain at least an annotation with data types definition.
+// The first row after annotations must be a line with column names.
+//
+// Annotated CSV example:
+//		#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,double,string,string,string,string
+//		#group,false,false,true,true,false,false,true,true,true,true
+//		#default,_result,,,,,,,,,
+//		,result,table,_start,_stop,_time,_value,_field,_measurement,location,sensor
+//		,,0,2020-02-17T22:19:49.747562847Z,2020-02-18T22:19:49.747562847Z,2020-02-18T10:34:08.135814545Z,23.4,tmp,air,livingroom,SHT31
+//		,,0,2020-02-17T22:19:49.747562847Z,2020-02-18T22:19:49.747562847Z,2020-02-18T22:08:44.850214724Z,21.6,tmp,air,livingroom,SHT31
+//
+// Each set of rows started with annotation is called "section".
+// CSV source can contain multiple sections, each section can have different columns (schema).
+package annotatedcsv
+
+import (
+	"encoding/base64"
+	"fmt"
+	"io"
+	"math"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/influxdata/influxdb-client-go/internal/csv"
+)
+
+type Column struct {
+	Name    string
+	Group   bool
+	Default interface{}
+	Type    string
+}
+
+// NewReader returns new Reader for parsing annotated csv stream.
+func NewReader(r io.ReadCloser) *Reader {
+	r1 := &Reader{
+		r:      csv.NewReader(r),
+		closer: r,
+	}
+	r1.r.FieldsPerRecord = -1
+	return r1
+}
+
+// Reader parses annotated csv stream with single or multiple sections.
+// Walking though the csv stream is done by repeatedly calling NextSection() and NextRow() until return false.
+// NextRow() can be also called initially, to advance straight to the first row of the first table.
+// Actual table schema (columns with names, data types, etc) is returned by Columns() method.
+// Data are acquired by Row() or by ValueByName() functions.
+// Preliminary end can be caused by an error, so when NextSection() or NextRow() return false, check Err() for an error.
+// Reader is automatically closed at the end reading or in case of an error.
+// Close() must be called manually in case of not reading stream till the end.
+type Reader struct {
+	cols []Column
+	row  []interface{}
+	err  error
+
+	hasPeeked bool
+	peekRow   []string
+	peekErr   error
+	closer    io.Closer
+	r         *csv.Reader
+	// columnIndexes maps column names to their indexes.
+	columnIndexes map[string]int
+}
+
+// NextSection advances to the next section in the csv stream and reports whether
+// there is one.
+// Any remaining data in the current section is discarded.
+// When there are no more sections, it returns false.
+func (r *Reader) NextSection() bool {
+	if r.err != nil {
+		return false
+	}
+	// Read all the current rows.
+	if r.cols != nil {
+		for r.NextRow() {
+		}
+	}
+	_, err := r.peek()
+	if err != nil {
+		r.err = err
+		return false
+	}
+	cols, err := r.readHeader()
+	if err != nil {
+		r.err = err
+		return false
+	}
+	r.cols = cols
+	return true
+}
+
+// Err returns any error encountered when parsing.
+func (r *Reader) Err() error {
+	if r.err == io.EOF {
+		return nil
+	}
+	return r.err
+}
+
+// Columns returns information on the columns in the current
+// section. It returns nil if there is no current section (for example
+// before NextSection has been called, or after NextSection returns false).
+func (r *Reader) Columns() []Column {
+	return r.cols
+}
+
+// NextRow advances to the next row in the current section.
+// If called in the beginning it also advances to the first section.
+// When there are no more rows in the current section, it returns false.
+func (r *Reader) NextRow() bool {
+	if r.err != nil {
+		return false
+	}
+	// Support for navigating straight to the first data row
+	if r.cols == nil && !r.NextSection() {
+		return false
+	}
+	row, err := r.readRow()
+	r.row = row
+	if row == nil {
+		r.err = err
+		r.cols = nil
+		return false
+	}
+	return true
+}
+
+// Row returns the values in the current row of the current section.
+// It returns nil if there is no current row.
+// All rows in a section have the same number of values.
+func (r *Reader) Row() []interface{} {
+	return r.row
+}
+
+// ValueByName returns value for given column name.
+// It returns nil if section has no value for such column.
+func (r *Reader) ValueByName(name string) interface{} {
+	if r.columnIndexes != nil {
+		if i, ok := r.columnIndexes[name]; ok {
+			return r.row[i]
+		}
+	}
+	return nil
+}
+
+// Close closes underlying reader.
+func (r *Reader) Close() error {
+	r.cols = nil
+	r.row = nil
+	r.columnIndexes = nil
+	return r.closer.Close()
+}
+
+func (r *Reader) safeClose() {
+	if err := r.Close(); err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "error closing csv reader: %v\n", err)
+	}
+}
+
+func (r *Reader) readRow() ([]interface{}, error) {
+	closer := func() {
+		r.safeClose()
+	}
+	defer func() { closer() }()
+	row, err := r.peek()
+	if err != nil {
+		return nil, err
+	}
+	if len(row) > 0 && strings.HasPrefix(row[0], "#") {
+		// Start of next table.
+		return nil, nil
+	}
+	_, _ = r.read()
+	if len(row)-1 != len(r.cols) {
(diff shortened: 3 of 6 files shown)


inluxclient/query.go
```diff
@@ -0,0 +1,315 @@
...
+	"encoding/csv"
+	"errors"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// TableColumn holds flux query table column properties
```
rogpeppe, line 14:
```suggestion
// TableColumn represents a column in a Flux query result table.
```

inluxclient/query.go
```diff
@@ -0,0 +1,315 @@
...
+type TableColumn struct {
+	Name         string
+	DataType     string
+	IsGroup      bool
+	DefaultValue string
+}
+
+// String returns TableColumn string dump
+func (f *TableColumn) String() string {
```
rogpeppe, line 23: I don't think this method pulls its weight - `Printf("%#v", col)` would produce something very similar.

inluxclient/query_test.go
```diff
@@ -0,0 +1,1083 @@
...
+	"io/ioutil"
+	"strings"
+	"testing"
+	"time"
+
+	influxclient "github.com/influxdata/influxdb-client-go/inluxclient"
+	"github.com/stretchr/testify/require"
+)
+
+func mustParseTime(s string) time.Time {
```
rogpeppe, line 16: FWIW I think it's nicer to have helper functions at the end of the file so they don't get in the way of top-down reading of the code.

inluxclient/query_test.go
```diff
@@ -0,0 +1,1083 @@
...
+,result,table,_start,_stop,_time,_value,_field,_measurement,a,b
+,,3,2020-02-17T22:19:49.747562847Z,2020-02-18T22:19:49.747562847Z,2020-02-18T22:08:44.62797864Z,0,i,test,0,adsfasdf
+,,3,2020-02-17T22:19:49.747562847Z,2020-02-18T22:19:49.747562847Z,2020-02-18T22:08:44.969100374Z,2,i,test,0,adsfasdf
+
+`
+	tablesMultiStructure := []expectedTable{
+		{ // Table 1
+			[]influxclient.TableColumn{
+				{DataType: "string", DefaultValue: "_result", Name: "result", IsGroup: false},
+				{DataType: "long", DefaultValue: "", Name: "table", IsGroup: false},
```
rogpeppe, line 156: You can omit the empty `DefaultValue` fields here.

inluxclient/query_test.go
```diff
@@ -0,0 +1,1083 @@
...
+	for _, table := range tables {
+		require.True(t, res.NextTable(), res.Err())
+		require.Nil(t, res.Err())
+		require.Equal(t, table.columns, res.Columns())
+		for _, row := range table.rows {
+			require.True(t, res.NextRow(), res.Err())
+			require.Nil(t, res.Err())
+			require.Equal(t, row, res.Values())
+			for i, c := range table.columns {
+				require.Equal(t, row[i], res.ValueByName(c.Name))
```
rogpeppe, line 42: I'm not that keen on `reflect.DeepEqual` being used to compare time values - it's generally considered to be bad practice.

You might consider using `github.com/frankban/quicktest` instead of testify - it makes it trivial to do this correctly (`qt.DeepEquals` uses the [cmp package](https://pkg.go.dev/github.com/google/go-cmp@v0.5.4/cmp) which understands `Equal` methods), and the API is a lot smaller.

inluxclient/query.go
```diff
@@ -0,0 +1,315 @@
...
+
+// QueryResults parses Flux query response stream.
+// Walking though the result set is done by repeatedly calling NextTable() and NextRow() until return false.
+// NextRow() can be also called initially, to advance straight to the first row of the first table.
+// Actual flux table schema (columns with names, data types, etc) is returned by Columns() method.
+// Data are acquired by Values() or by ValueByName() functions.
+// Preliminary end can be caused by an error, so when NextTable() or NextRow() return false, check Err() for an error.
+// Reader is automatically closed at the end reading or in case of an error.
+type QueryResults struct {
+	io.Closer
```
rogpeppe, line 63: I think I'd avoid embedding this in favour of an explicit `Close` method - that way we have the freedom to change instance variables on close if we want, and we don't really want it to be public.

inluxclient/query.go
```diff
@@ -0,0 +1,315 @@
...
+	csvReader *csv.Reader
+	columns   []TableColumn
+	values    []interface{}
+	err       error
+	lastRow   []string
+	names     map[string]int
```
rogpeppe, line 69:
```suggestion
        // columnIndexes maps column names to their indexes.
	columnIndexes     map[string]int
```
I think it's nicer to name the map after the values rather than the keys.

inluxclient/query.go
```diff
@@ -0,0 +1,315 @@
...
+// Data are acquired by Values() or by ValueByName() functions.
+// Preliminary end can be caused by an error, so when NextTable() or NextRow() return false, check Err() for an error.
+// Reader is automatically closed at the end reading or in case of an error.
+type QueryResults struct {
+	io.Closer
+	csvReader *csv.Reader
+	columns   []TableColumn
+	values    []interface{}
+	err       error
+	lastRow   []string
```
rogpeppe, line 68: I find "last" problematic as a name because of the English language ambiguity between "the very last" and "the most recent". Maybe `currentRow` would be better here?

rogpeppe (commented):
This is going in a good direction, thanks! It's great to have some good testing coverage. I have reservations about the approach used in the table reading code and have a suggestion for an alternative. I wonder what you think.

inluxclient/query.go
```diff
@@ -0,0 +1,315 @@
...
+	DefaultValue string
+}
+
+// String returns TableColumn string dump
+func (f *TableColumn) String() string {
+	return fmt.Sprintf("{name: %s, datatype: %s, defaultValue: %s, group: %v}", f.Name, f.DataType, f.DefaultValue, f.IsGroup)
+}
+
+const (
+	stringDatatype       = "string"
```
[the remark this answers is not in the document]
vlastahajek: These constants will be used also in the `Decode()` function, to validate type compatibility.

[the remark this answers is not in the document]
rogpeppe: That's quite possible, but the Decode function doesn't exist yet, so it's hard to be sure if actually will be necessary (I can  imagine an implementation where there's still only a single switch statement - if fact I suspect it might be a good idea to implement toValue in terms of Decode, but we'll have to see about that. I'd suggest keeping the code as simple as possible at each step and refactoring when needed. That way it's easy to review and understand each PR on its own terms without guessing the future.
Of course there might be times when we have to look ahead a bit in the implementation, but I'm not sure this is one of those.

inluxclient/query.go
```diff
@@ -0,0 +1,315 @@
...
+	} else {
+		// test columns consistency in case of data row or already discovered annotations
+		if (row[0] == "" || parsingState == parsingStateAnnotation) && len(row)-1 != len(r.columns) {
+			r.err = fmt.Errorf("parsing error, row has different number of columns than the table: %d vs %d", len(row)-1, len(r.columns))
+			return false
+		}
+	}
+	switch {
+	case row[0] == "":
+		switch parsingState {
```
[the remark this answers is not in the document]
vlastahajek: Having a separate library for decoding Flux CSV is definitely a good idea. 👍 
 I know some users of the current Go client use the `QueryTableResult` as a Flux parser to read their Flux exports.
 
 Used algorithms and patterns is always a subjective approach. Your `annotatedcsv.Reader`  parser looks indeed more simple by separating parsing of header and row. If the `Decode` is suitable for the reader (I don't see why not), it can replace the current `QueryResults`.
 
 What would be the next step here? 
 Should I wait for you to finish it or should I contribute to the package by adding test, fixes, etc to help you to finish it?
 Will it be later moved to the influxdata GitHub org?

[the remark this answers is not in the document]
rogpeppe: > What would be the next step here?
> Should I wait for you to finish it or should I contribute to the package by adding test, fixes, etc to help you to finish it?
> Will it be later moved to the influxdata GitHub org?

Would you be up for taking my code above and writing tests/fixes to bring it up to scratch? If not, np, I can do it, but I'm hoping it's understandable enough that I don't have to. It should be possible to re-purpose your existing tests, I think, which should save some time.

I think it's fine to have it as a sub-package within this module, as there are some influx-specific things in it (e.g. the `tag` type).

inluxclient/query_test.go
vlastahajek: Sure, but with it, there is explicitly shown I want to verify the empty string for the field.

inluxclient/query_test.go
vlastahajek: Ok, I will fix comparing time values. 
Generally, `testify` seems more friendly and straightforward to use to me.

inluxclient/query.go
vlastahajek: Public `Close()` can be useful when Flux CSV is not read till the end to close the HTTP response body.

inluxclient/query.go
vlastahajek: Ok

inluxclient/query.go
rogpeppe: I'm not suggesting that you don't have a public `Close` method, but that it's defined explicitly rather than via embedding:
```
func (q *QueryResults) Close() error {
    return q.r.Close()
}
```

That way we can easily attach arbitrary logic to the Close method if needed, and there's no worry that a caller might change the `Closer` field of the struct.

vlastahajek: Ok

inluxclient/query_test.go
vlastahajek: Ok

inluxclient/query.go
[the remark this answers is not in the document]
vlastahajek: I have the `Decode` in progress in [my fork](https://github.com/bonitoo-io/influxdb-client-go/blob/tmp/query_cont/inluxclient/query_decode.go). It is far from finishing, but if it is going in the wrong direction,  please, let me know.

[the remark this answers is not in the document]
rogpeppe: Thanks for pointing me to that. I had a think about it, and I think I'd choose a slightly different direction (sorry to be a pain again!). Two main things:
- I think that rather than iterating over all the fields of the type, it might be better to iterate over the known columns instead (on the basis that if you're decoding into a struct with fields that aren't mentioned in the columns, that probably doesn't matter).
- It's usually worth trying to avoid doing more type reflection work than necessary - it would be nice to make this reasonably efficient for tables with many rows.

The second point is conventionally addressed by creating a closure to do the work for a given known type, with as many aspects as possible factored out of the inner part of the function. In many places (e.g. encoding/json), that's done with a global `sync.Map` variable keyed by type, but in this case, I think we might be able to assume that tables have enough rows that just caching the converter function for the latest known type and set of columns is enough.

I did a very rough and rapid sketch of how I think the structure of the decoder might look in this style:

```
type Reader struct {
	// decodeType holds the type that was last decoded into.
	// This is reset at the start of each section.
	decodeType reflect.Type
	// colSetters holds an element for each column in the current
	// section. Given an instance of decodeType, it decodes
	// the column string value into the appropriate field.
	colSetters []func(reflect.Value, string) error
}

// initColumns initializes r.colSetters for the given columns.
// It's called once per section.
func (r *Reader) initColumns(t reflect.Type, columns []columnInfo) error {
	if t == r.decodeType {
		return nil
	}
	if t.Kind() != reflect.Ptr {
		return fmt.Errorf("cannot decode into non-pointer value")
	}
	t = t.Elem()
	if t.Kind() != reflect.Struct {
		return fmt.Errorf("cannot decode into non-struct value")
	}
	setters := make([]fieldSetter, len(columns))
	
	for i, col := range columns {
		find corresponding field in struct
		fkind := fieldKind(fieldType)
		convert, ok := conversions[conv{columnType, fkind]
		if !ok {
			error('cannot convert from column type %v to %v", fieldType, columnType)
		}
		setters[i] = func(v reflect.Value, s string) error {
			fv := v.FieldByIndex(fieldIndex)
			return convert(fv, s)
		}
	}
	r.decodeType = t
	return nil
}

// conv represents a possible conversion source and destination.
type conv struct {
	from columnKind
	to reflect.Kind
}

// fieldKind represents one of the possible kinds of struct field.
// It's similar to reflect.Kind (every reflect.Kind constant i
// is represented as fieldKind(i)) except that it also has
// defined constants for field types that have special treatment,
// such as time.Duration.
type fieldKind uint

var intKinds = []fieldKind{
	fieldKind(reflect.Int),
	fieldKind(reflect.Int8),
	fieldKind(reflect.Int16),
	etc,
}

const (
	durationKind = fieldKind(255 + iota)
	timeKind
	bytesKind
)

type colType uint

const (
	stringCol = colType(iota)
	durationCol
	longCol
	etc
)

type fieldSetter = func(v reflect.Value, s string) error

var conversions = func() map[conv] fieldSetter {
	m := make(map[conv] fieldSetter)
	for _, k := range intKinds {
		m[conv{longCol, k}] = toInt
		m[conv{uLongCol, k}] = toInt
	}
	for _, k := range uintKinds {
		m[conv{uLongCol, k}] = toUint
	}
	for _, k := range floatKinds {
		m[conv{longCol, k}] = toFloat
		m[conv{uLongCol, k}] = toFloat
		m[conv{doubleCol, k}] = toFloat
	}
	m[conv{stringCol, reflect.String}] = toString
	m[conv{durationCol, durationKind}] = toDuration
	m[conv{durationCol, reflect.String}] = toString
	m[conv{base64BinaryCol, bytesKind}] = toBytes
	m[conv{timeColRFC, timeKind}] = toTime
	m[conv{timeColRFCNano, timeKind}] = toTime
	
	for col, t := range canonicalKinds {
		m[conv{col, reflect.Interface}] = toInterface(col)
	}
}()

// canonicalTypes holds the Go type that best represents
// all the column kinds (keyed  by column type).
var canonicalTypes = []reflect.Type {
	stringCol: reflect.TypeOf(""),
	durationCol: reflect.TypeOf(time.Duration),
	longCol: reflect.TypeOf(int64(0)),
	uLongCol: reflect.TypeOf(uint64(0))
	doubleCol: reflect.TypeOf(0.0),
	timeColRFC: reflect.TypeOf(time.Time{}),
	timeColRFCNano: reflect.TypeOf(time.Time{}),
}

var basicKinds = map[reflect.Kind] bool {
	reflect.Int,
	reflect.Int16,
	...
	reflect.Uint,
	...
	reflect.

func fieldKindOf(t reflect.Type) fieldKind {
	switch t {
	case time.Duration:
		return durationKind
	case time.Time:
		return timeKind
	case reflect.Slice:
		if reflect.Elem().Kind() == reflect.Uint8 {
			return bytesKind
		}
	}
	return fieldKind(t.Kind())
}

func toInterface(col colType) fieldSetter {
	t := canonicalKinds[col]
	convert, ok := conversions[conv{col, fieldKindOf(t)]
	if !ok {
		panic("conversion not found (should not happen)")
	}
	return func(v reflect.Value, s string) error {
		e := reflect.New(t).Elem()
		if err := convert(e, s); err != nil {
			return err
		}
		v.Set(e)
		return nil
	}
}

func toString(v reflect.Value, s string) error {
	v.SetString(s)
	return nil
}

func toFloat(v reflect.Value, s string) error {
	x, err := strconv.ParseFloat(s, 64, 10)
	if err != nil {
		return err
	}
	if v.OverflowFloat(x) {
		return fmt.Errorf("overflow")
	}
	v.SetFloat(x)
	return nil
}

func toBool(v reflect.Value, s string) error {
	var b bool
	switch s {
	case "true":
		b = true
	case "false"
		b = false
	default:
		return fmt.Errorf("invalid bool value")
	}
	v.SetBool(b)
	return nil
}

func toInt(v reflect.Value, s string) error {
	x, err := strconv.ParseInt(s, 64, 10)
	if err != nil {
		return err
	}
	if v.OverflowInt(x) {
		return fmt.Errorf("overflow")
	}
	v.SetInt(x)
	return nil
}
```

Note: Instead of having string constants for the column types, I've defined a numeric type - there would be some code somewhere to convert to that via a switch or a map lookup.

inluxclient/query.go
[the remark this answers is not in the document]
vlastahajek: _Conversation continued via emails_

PR #237, opened 2021-03-18
fix: don't attempt retries if MaxRetries is set to zero

### LINKED ISSUE #236: Retries are attempted even if MaxRetries is set to zero
__Steps to reproduce:__
Given the following client code snippet

```
	options := influxdb2.DefaultOptions().SetMaxRetries(0)
	client := influxdb2.NewClientWithOptions(url, authToken, options)
	defer client.Close()
	writeAPI := client.WriteAPIBlocking("", dbName)
        err := writeAPI.WritePoint(context.Background(), point)
        if err != nil {
            fmt.Printf("Error: %v", err)
        }
```

where `url` is the URL of an InfluxDB instance that is unreachable (i.e. results in a connection error), the `WritePoint()` method may attempt an implicit deferred retry (instead of immediately returning a connection error), even though MaxRetries was set to 0.

__Expected behavior:__
The `WritePoint()` method should always return an error in this case.

__Actual behavior:__
No error is returned.

__Specifications:__
 - Client Version: 2.2.2
 - InfluxDB Version: n/a (testing against an unreachable InfluxDB instance)
 - Platform: MacOS 10.15.7

DaveHasteSpringer:
Closes #236 

## Proposed Changes

When a write error occurs, explicitly test for `MaxRetries()` being zero, and do not proceed with retry logic if this is the case.

## Checklist

- [ ] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x] Tests pass
- [ ] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/internal/write/service.go
+++ b/internal/write/service.go
@@ -99,7 +99,7 @@ func (w *Service) HandleWrite(ctx context.Context, batch *Batch) error {
 		if batchToWrite != nil {
 			perror := w.WriteBatch(ctx, batchToWrite)
 			if perror != nil {
-				if perror.StatusCode == 0 || perror.StatusCode >= http.StatusTooManyRequests {
+				if w.writeOptions.MaxRetries() != 0 && (perror.StatusCode == 0 || perror.StatusCode >= http.StatusTooManyRequests) {
 					log.Errorf("Write error: %s\nBatch kept for retrying\n", perror.Error())
 					if perror.RetryAfter > 0 {
 						batchToWrite.retryDelay = perror.RetryAfter * 1000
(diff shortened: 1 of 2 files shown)


vlastahajek:
@DaveHasteSpringer, thanks for the PR! This definitely makes sense.
Please, fix the last commit message.

DaveHasteSpringer:
Thanks @vlastahajek, I've squashed the two commits into one (which makes more sense as the unit test goes hand-in-hand with the change made), so hopefully the commit message should be OK now.

vlastahajek (approved):
Thanks!

PR #239, opened 2021-03-24
fix: Removing retrying from blocking write

### LINKED ISSUE #235: WriteAPIBlocking.WritePoint does not always return error if influx not reachable
Hi,

I noticed that the WriteAPIBlocking gives inconsistent return values when InfluxDB is not reachable.
When testing out the app behavior when Influx is not running, I noticed that some write attempts yield errors but many run through as if they worked fine (err==nil). 

In my case I'm writing concurrently from 3 separate go routines (about 3 points each per second) calling a function with the code below. According to issues/docs the WriteAPIBlocking should be thread-safe. 

Any ideas what could be the issue?

The code in questions is based on the basic blocking write example

```
	        startTime := time.Now()
		err := influx.writeAPIBlocking.WritePoint(context.Background(), newPoints...)
		if err != nil {
			logger.Errorf("Error saving data to InfluxDB. Reason: %s", err.Error())
			return err
		}
		log.Infof("Saved %d %s point(s) in %d ms", len(newPoints), dataset.MeasurementName, time.Since(startTime).Milliseconds())
		return nil
```

__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. Turn off influx db (or use wrong url/port)
2. Attempt to write points (from several go routines)
3. err struct is sometimes nil indicating that everything was written OK

__Expected behavior:__
a not reachable influxdb should consistently produce errors on write

__Actual behavior:__
err is sometimes nil

__Specifications:__
 - Client Version: 2.2.2
 - InfluxDB Version: 2.0.4
 - Platform: Ubuntu 18.04 x64

### LINKED ISSUE #236: Retries are attempted even if MaxRetries is set to zero

vlastahajek:
Closes #235

## Proposed Changes

Blocking write client doesn't use retry handling.

## Checklist

- [X] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass
- [X] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/api/http/error.go
+++ b/api/http/error.go
@@ -5,6 +5,7 @@
 package http
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 )
@@ -28,6 +29,10 @@ func (e *Error) Error() string {
 	default:
 		return "Unexpected status code " + strconv.Itoa(e.StatusCode)
 	}
+}
+
+func (e *Error) Unwrap() error {
+	return errors.New(e.Error())
 }
 
 // NewError returns newly created Error initialised with nested error and default values
--- a/internal/write/service.go
+++ b/internal/write/service.go
@@ -55,7 +55,15 @@ func NewService(org string, bucket string, httpService http2.Service, options *w
 	if retryBufferLimit == 0 {
 		retryBufferLimit = 1
 	}
-	return &Service{org: org, bucket: bucket, httpService: httpService, writeOptions: options, retryQueue: newQueue(int(retryBufferLimit)), retryExponentialBase: 5}
+	u, _ := url.Parse(httpService.ServerAPIURL())
+	u, _ = u.Parse("write")
+	params := u.Query()
+	params.Set("org", org)
+	params.Set("bucket", bucket)
+	params.Set("precision", precisionToString(options.Precision()))
+	u.RawQuery = params.Encode()
+	writeURL := u.String()
+	return &Service{org: org, bucket: bucket, httpService: httpService, url: writeURL, writeOptions: options, retryQueue: newQueue(int(retryBufferLimit)), retryExponentialBase: 5}
 }
 
 func (w *Service) HandleWrite(ctx context.Context, batch *Batch) error {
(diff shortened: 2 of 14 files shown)


api/http/error.go
```diff
@@ -30,6 +31,10 @@ func (e *Error) Error() string {
 	}
 }
 
+func (e *Error) ToError() error {
```
sranka, line 34:
```suggestion
func (e *Error) Unwrap() error {
```
Unwrap was introduced since go 1.13

internal/write/service.go
```diff
@@ -55,7 +55,15 @@ func NewService(org string, bucket string, httpService http2.Service, options *w
 	if retryBufferLimit == 0 {
 		retryBufferLimit = 1
 	}
-	return &Service{org: org, bucket: bucket, httpService: httpService, writeOptions: options, retryQueue: newQueue(int(retryBufferLimit)), retryExponentialBase: 5}
+	u, _ := url.Parse(httpService.ServerAPIURL())
+	u, _ = u.Parse("write")
+	params := u.Query()
+	params.Set("org", org)
+	params.Set("bucket", bucket)
+	params.Set("precision", precisionToString(options.Precision()))
```
sranka, line 63: this is probably related to a change in the WriteURL func

sranka (changes requested):

api/http/error.go
vlastahajek: Unwrap seems to be related to the case when this `Error` struct just wraps another `error`. This function just creates `error` from the string composed from wrapped fields.

internal/write/service.go
vlastahajek: Lazy setting the complete write url was removed as considered useless and complicated as it needed to be synced.

api/http/error.go
sranka: `ToError` func on an `Error` struct that itself implements =error= is still confusing to me. I think think that Unwrap would be better.

sranka (approved):

PR #240, opened 2021-04-02
feat(Query): Adding Reader.Decode

vlastahajek:
## Proposed Changes
- Added `Reader.Decode(x interface{}) errror` to convert CSV values to a struct fields or a slice
- `Reader.NextRow()` doesn't convert values 
- `Reader.Row()` returns raw CSV data

## Checklist

- [ ] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass
- [X] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/annotatedcsv/decode.go
+++ b/annotatedcsv/decode.go
@@ -1,0 +1,459 @@
+// Copyright 2021 InfluxData, Inc. All rights reserved.
+// Use of this source code is governed by MIT
+// license that can be found in the LICENSE file.
+
+package annotatedcsv
+
+import (
+	"encoding/base64"
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+	"time"
+
+	ireflect "github.com/influxdata/influxdb-client-go/internal/reflect"
+)
+
+// Decode decodes the current row into x, which should be
+// a pointer to a struct or a pointer to a slice.
+//
+// When it's a pointer to a struct, columns in the row are decoded into
+// appropriate fields in the struct, using the similar tag conventions
+// described by encoding/json to determine how to map
+// column names to struct fields. Tag prefix must start with "csv:":
+//
+//  type Point struct {
+//      Timestamp  time.Time `csv:"_time"`
+//      Value      float32   `csv:"_value"`
+//      Location   string    `csv:"location"`
+//      Sensor     string    `csv:"type"`
+//  }
+//
+//  var p Point
+//  err := r.Decode(&p)
+//
+// When it's a pointer to a slice, the slice is changed
+// to have one element for each column (reusing
+// space in the slice if possible), and each element is
+// set to the value in the corresponding column.
+//
+// When decoding into an empty interface value, the resulting
+// type depends on the column type:
+//
+// - string, tag or unrecognized: string
+// - double: float64
+// - unsignedLong: uint64
+// - long: int64
+// - boolean: bool
+// - duration: time.Duration
+// - dateTime: time.Time
+//
+// Any value can be decoded into a string without
+// error - the result is the value in the CSV, so
+//
+//     var row []string
+//     r.Decode(&row)
+//
+// will always succeed and provide all the values in the column as strings.
+//
+// Similarly, any row can be decoded into []interface{}, which will convert
+// all values into elements of the slice according to the rules for empty interfaces above.
+//
+// Slice capacity will be reused when possible.
+func (r *Reader) Decode(x interface{}) error {
+	t := reflect.TypeOf(x)
+	if err := r.initColumns(t, r.cols); err != nil {
+		return err
+	}
+	return r.decodeRow(reflect.ValueOf(x), r.row)
+}
+
+// initColumns initializes r.colSetters for decoding the given columns
+// into a value of the given type.
+func (r *Reader) initColumns(t reflect.Type, columns []Column) error {
+	if t == r.decodeType {
+		return nil
+	}
+	if t.Kind() != reflect.Ptr {
+		return fmt.Errorf("cannot decode into non-pointer type %v", t)
+	}
+	et := t.Elem()
+	if et.Kind() != reflect.Struct && et.Kind() != reflect.Slice {
+		return fmt.Errorf("cannot decode into %v", t)
+	}
+	setters := make([]fieldSetter, len(columns))
+	switch et.Kind() {
+	case reflect.Struct:
+		fieldsMap := make(map[string]reflect.StructField)
+		fields := ireflect.VisibleFields(et)
+		for _, f := range fields {
+			name := f.Name
+			if tag, ok := f.Tag.Lookup("csv"); ok {
+				if tag == "-" {
+					continue
+				}
+				if i := strings.Index(tag, ","); i >= 0 {
+					if i != len(tag)-1 {
+						return fmt.Errorf("tag attributes are not supported")
+					}
+					tag = tag[:i]
+				}
+				name = tag
+			}
+			fieldsMap[name] = f
+		}
+
+		for i, col := range columns {
+			f, ok := fieldsMap[col.Name]
+			if !ok {
+				// The column isn't mentioned in the struct.
+				continue
+			}
+			ftype := f.Type
+			colType, ok := columnTypes[col.Type]
+			if !ok {
+				// ignore invalid type and use string
+				colType = stringCol
+			}
+			convert, ok := conversions[conv{colType, fieldKindOf(ftype)}]
+			if !ok {
+				return fmt.Errorf("cannot convert from column type %s to %v", col.Type, ftype)
+			}
+			setters[i] = func(v reflect.Value, colIndex int) error {
+				return r.convertColumnValue(v.FieldByIndex(f.Index), colIndex, convert)
+			}
+		}
+		r.decodeRow = func(v reflect.Value, row []string) error {
+			if v.IsNil() {
+				return fmt.Errorf("decode into nil %s", v.Type())
+			}
+			v = v.Elem()
+			for i, c := range r.colSetters {
+				if c == nil {
+					continue
+				}
+				if err := c(v, i); err != nil {
+					return err
+				}
+			}
+			return nil
+		}
+	case reflect.Slice:
+		s := et.Elem()
+		if s != stringType && s != emptyInterfaceType {
+			return fmt.Errorf("cannot decode into *[]%v", s)
+		}
+		fkind := fieldKindOf(s)
+		for i, col := range columns {
+			colType, ok := columnTypes[col.Type]
+			if !ok {
+				// ignore invalid type and use string
+				colType = stringCol
+			}
+
+			convert, ok := conversions[conv{colType, fkind}]
+			if !ok {
+				return fmt.Errorf("cannot convert from column type %s to %v", col.Type, s)
+			}
+			setters[i] = func(v reflect.Value, colIndex int) error {
+				if err := r.convertColumnValue(v.Index(colIndex), colIndex, convert); err != nil {
+					return err
+				}
+				return nil
+			}
+		}
+		r.decodeRow = func(v reflect.Value, row []string) error {
+			if v.IsNil() {
+				return fmt.Errorf("decode into nil %s", v.Type())
+			}
+			e := v.Elem()
+			c := len(r.colSetters)
+			if e.Cap() < c {
+				e = reflect.MakeSlice(e.Type(), c, c)
+			} else {
+				e = e.Slice(0, c)
+			}
+			for i, setf := range r.colSetters {
+				if err := setf(e, i); err != nil {
+					return err
+				}
+			}
+			v.Elem().Set(e)
+			return nil
+		}
+	}
+
+	r.colSetters = setters
+	r.decodeType = t
+
+	return nil
+}
+
+// convertColumnValue set a value from current row of given column index
+// to a struct or a slice field value
+func (r *Reader) convertColumnValue(v reflect.Value, colIndex int, convert valueSetter) error {
+	if err := convert(v, r.row[colIndex]); err != nil {
+		return fmt.Errorf(`cannot convert value %q in column of type %q to Go type %v at line %d: %w`, r.row[colIndex], r.cols[colIndex].Type, v.Type(), r.r.Line(), err)
+	}
+	return nil
+}
+
+// fieldKind represents one of the possible kinds of struct field.
+// It's similar to reflect.Kind (every reflect.Kind constant i
+// is represented as fieldKind(i)) except that it also has
+// defined constants for field types that have special treatment,
+// such as time.Duration.
+type fieldKind uint
+
+// conv represents a possible conversion source and destination.
+type conv struct {
+	from colType
+	to   fieldKind
+}
+
+// intKinds holds all supported signed integer kinds
+var intKinds = []fieldKind{
+	fieldKind(reflect.Int),
+	fieldKind(reflect.Int8),
+	fieldKind(reflect.Int16),
+	fieldKind(reflect.Int32),
+	fieldKind(reflect.Int64),
+}
+
+// uintKinds holds all supported unsigned integer kinds
+var uintKinds = []fieldKind{
+	fieldKind(reflect.Uint),
+	fieldKind(reflect.Uint8),
+	fieldKind(reflect.Uint16),
+	fieldKind(reflect.Uint32),
+	fieldKind(reflect.Uint64),
+}
+
+// floatKinds holds all supported floating point number kinds
+var floatKinds = []fieldKind{
+	fieldKind(reflect.Float32),
+	fieldKind(reflect.Float64),
+}
+
+// rest of supported type kinds
+const (
+	durationKind = fieldKind(255 + iota)
+	timeKind
+	bytesKind
+)
+
+// colType represents the type of an annotated CSV column.
+type colType uint
+
+const (
+	stringCol = colType(iota)
+	boolCol
+	durationCol
+	longCol
+	unsignedLongCol
+	doubleCol
+	base64BinaryCol
+	rfc3339Col
+)
+
+// columnTypes maps annotated CSV types
+// to integer column type
+var columnTypes = map[string]colType{
+	"string":               stringCol,
+	"boolean":              boolCol,
+	"duration":             durationCol,
+	"long":                 longCol,
+	"unsignedLong":         unsignedLongCol,
+	"double":               doubleCol,
+	"base64Binary":         base64BinaryCol,
+	"dateTime:RFC3339":     rfc3339Col,
+	"dateTime:RFC3339Nano": rfc3339Col,
+	"dateTime":             rfc3339Col,
+}
+
+// conversions maps all possible conversions from column types to field kinds
+var conversions map[conv]valueSetter
+
+// stringType is the exact type for the empty string
+var stringType = reflect.TypeOf("")
+
+// emptyInterfaceType is the exact type for the empty interface{}
+var emptyInterfaceType = reflect.TypeOf(new(interface{})).Elem()
+
+// canonicalTypes holds the Go type that best represents
+// all the annotated CSV column kinds (keyed  by column type).
+var canonicalTypes = []reflect.Type{
+	stringCol:       reflect.TypeOf(""),
+	boolCol:         reflect.TypeOf(true),
+	durationCol:     reflect.TypeOf(time.Duration(1)),
+	longCol:         reflect.TypeOf(int64(0)),
+	unsignedLongCol: reflect.TypeOf(uint64(0)),
+	doubleCol:       reflect.TypeOf(0.0),
+	base64BinaryCol: reflect.TypeOf([]byte{}),
+	rfc3339Col:      reflect.TypeOf(time.Time{}),
+}
+
+// fieldKindOf determines a fieldKind for given reflect.Type
+func fieldKindOf(t reflect.Type) fieldKind {
+	switch t {
+	case canonicalTypes[durationCol]:
+		return durationKind
+	case canonicalTypes[rfc3339Col]:
+		return timeKind
+	case canonicalTypes[base64BinaryCol]:
+		return bytesKind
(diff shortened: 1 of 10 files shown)


annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+//     r.Decode(&row)
+//
+// will always succeed and provide all the values in the column as strings.
+func (r *Reader) Decode(x interface{}) error {
+	t := reflect.TypeOf(x)
+	if err := r.initColumns(t, r.cols); err != nil {
+		return err
+	}
+	et := t.Elem()
+	switch et.Kind() {
```
rogpeppe, line 64: I don't think you should need this switch. I think you should be able to arrange things inside `initColumns` such that after it's run, there's just a single function inside the reader that can be called to do all the decoding for a row.

e.g. (in `Reader`)

```
    // decodeRow decodes the row with the given values into v,
    // which must have type decodeType.
    decodeRow func(v reflect.Value, row []string) error
```

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+
+	return nil
+}
+
+// convertColumnValue set a value from current row of given column index
+// to a struct or a slice field value
+func (r *Reader) convertColumnValue(v reflect.Value, colIndex int) error {
+	s := stringTernary(r.row[colIndex], r.cols[colIndex].Default)
+	if err := r.colSetters[colIndex](v, s); err != nil {
+		return fmt.Errorf(`cannot convert value "%s" to type "%s" at line %d: %w`, s, r.cols[colIndex].Type, r.r.Line(), err)
```
rogpeppe, line 151:
```suggestion
		return fmt.Errorf(`cannot convert value %q in column of type %q to Go type %v at line %d: %w`, s, r.cols[colIndex].Type, v.Type(), r.r.Line(), err)
```

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+	case canonicalTypes[durationCol]:
+		return durationKind
+	case canonicalTypes[timeColRFC], canonicalTypes[timeColRFCNano]:
+		return timeKind
+	case canonicalTypes[base64BinaryCol]:
+		if t.Elem().Kind() == reflect.Uint8 {
+			return bytesKind
+		}
+	}
+	return fieldKind(t.Kind())
```
rogpeppe, line 279: This isn't quite right for interface types. We shouldn't be allowing non-empty interface types, but this logic conflates all interface types together, which means that `toInterface` will panic for non-empty interfaces.

```suggestion
        if t.Kind() == reflect.Interface && t != emptyInterfaceType {
                // Only the empty interface type is allowed.
                return reflect.Invalid
         }
	return fieldKind(t.Kind())
```

Let's add a test case for a non-empty interface too.

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+
+// stringTernary returns second argument of first is empty, otherwise first
+func stringTernary(s string, d string) string {
+	if s != "" {
+		return s
+	}
+	return d
+}
+
+func init() {
```
rogpeppe, line 426: FWIW I tend to prefer to avoid `init` functions like this in favour of a function initializer style:

```
var conversions = func() map[conv] fieldSetter {
    etc
}()
```

That way you never have to worry about initialization order.

annotatedcsv/decode_test.go
```diff
@@ -0,0 +1,844 @@
...
+	s1 := &struct {
+		S []byte
+		D time.Duration
+		T time.Time
+		I int8
+		U uint8
+		B bool
+		F float32
+	}{}
+	require.Error(t, toBytes(reflect.ValueOf(s1).Elem().FieldByName("S"), "#"))
```
rogpeppe, line 600: I'd much prefer it if all these tests were done through the public API, please.

That way the tests give us full confidence that the overall logic works, and we have freedom to refactor internally without needing to change the tests.

I'd also like to see the actual error string in the tests. Too many times I've found tests that are testing for just the existence of an error but the tests were passing with an actual error that is either unreadable or misleading or just not the error that's expected.

A lot of this looks like it would fit a table-driven test very well:

```
var singleRowDecodeTests = struct{
    testName string
    csv string
    value interface{}
    expectError string
}{{
   testName: "Simple",
   csv: `#datatype,long,double,dateTime,string
#default,,,,
,index,score,time,name
,0,3.3,2021-02-18T10:34:08.135814545Z,Thomas
`,
    value: struct{
         Score int `flux:"score"`
         Name string `flux:"name"`
    }{
        Score: 3,
        Name: "Thomas",
    },
}, {
    testName: "InvalidDuration",
    csv: `#datatype,duration
,d
,#
`,
     value: struct{
         D time.Duration `flux:"d"`
     },
     expectError: "cannot convert column d of type duration to time.Duration: invalid etc",
}}

func TestSingleRowDecode(t *testing.T) {
    for _, test := range singleRowDecodeTests {
         t.Run(test.testName, func(t *testing.T) {
             decVal := reflect.New(reflect.TypeOf(test.value))
             r := NewReader(strings.NewReader(test.csv))
             err := r.Decode(decVal.Interface())
             if test.expectError {
                 require.Error(t, err)
                 require.Equal(t, test.expectError, err.Error()
                 return
              }
              require.NoError(t, err)
              require.Equal(t, test.value, decVal.Elem().Interface())
         })
    }
}
```

You can probably cover a great deal of the functionality by populating that table, leaving other cases like reading multiple rows and different decode types for other more specific tests.

To reduce verbosity, a couple of helper functions might turn out useful (e.g. a function that returns CSV data with a single row with a given field type, field name and field value; a function that returns a struct containing a single field with a given type, value and flux tag).

annotatedcsv/decode_test.go
```diff
@@ -0,0 +1,844 @@
...
+	s2 := &struct {
+		Index int `flux:"index"`
+	}{}
+	err = res.Decode(s2)
+	require.Error(t, err)
+	assert.Equal(t, `cannot convert value "1.0" to type "long" at line 4: strconv.ParseInt: parsing "1.0": invalid syntax`, err.Error())
+
+}
+
+// MustParseTime returns  parsed dateTime in RFC3339 and panics if it fails
```
rogpeppe, line 837:
```suggestion
// mustParseTime parses s as an RFC3339 timestamp and panics on error.
```

annotatedcsv/reader.go
```diff
@@ -107,6 +109,7 @@ func (r *Reader) NextSection() bool {
 		return false
 	}
 	r.cols = cols
+	r.decodeType = nil
```
rogpeppe, line 112:
```suggestion
        // Zero the cached decode type because it probably won't be
        // valid for the next section.
	r.decodeType = nil
```

annotatedcsv/reader.go
```diff
@@ -142,25 +145,14 @@ func (r *Reader) NextRow() bool {
 	return true
 }
 
-// Row returns the values in the current row of the current section.
+// Row returns the raw values in the current row of the current section.
```
rogpeppe, line 148: I wonder if this should fill in default values rather than returning them empty.
WDYT?

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+
+// colType is integer based representation annotated CSV types
+type colType uint
+
+const (
+	stringCol = colType(iota)
+	boolCol
+	durationCol
+	longCol
+	uLongCol
```
rogpeppe, line 228: Let's make this reflect the original naming, even if it is a tad verbose:

```suggestion
	unsignedLongCol
```

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+	doubleCol
+	base64BinaryCol
+	timeColRFC
+	timeColRFCNano
```
rogpeppe, line 232: It would be good to be consistent about the naming of these constants and always put `Col` at the end.
```suggestion
	rfc3339Col,
	rfc3339NanoCol
```

Also, I think you should be able to make do with only a single time kind here, because the Go time package will always let you parse RFC3339Nano with RFC3339.

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+// Decode decodes the current row into x, which should be
+// a pointer to a struct or a pointer to a slice.
+//
+// When it's a pointer to a struct, columns in the row are decoded into
+// appropriate fields in the struct, using the similar tag conventions
+// described by encoding/json to determine how to map
+// column names to struct fields. Tag prefix must start with "flux:":
+//
+//  type Point struct {
+//      Timestamp  time.Time `flux:"_time"`
```
rogpeppe, line 26: I'm wondering if `flux` is the correct annotation here, and thinking about `csv` instead, as we're really about CSV here, and flux might have several possible output formats in the future. WDYT?

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+// - duration: time.Duration
+// - dateTime: time.Time
+//
+// Any value can be decoded into a string without
+// error - the result is the value in the CSV, so
+//
+//     var row []string
+//     r.Decode(&row)
+//
+// will always succeed and provide all the values in the column as strings.
```
rogpeppe, line 57:
```suggestion
// will always succeed and provide all the values in the column as strings.
//
// Similarly, any row can be decoded into []interface{}, which will convert
// all values into elements of the slice according to the rules for empty interfaces above.
//
// Slice capacity will be reused when possible.
```

rogpeppe (changes requested):
Thanks very much for this! Lots of things are right, but I have a bunch of comments and suggestions as usual :)

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+		if t.Elem().Kind() == reflect.Uint8 {
+			return bytesKind
+		}
+	}
+	return fieldKind(t.Kind())
+}
+
+type fieldSetter = func(v reflect.Value, s string) error
+
+type fieldTypeOf func(col Column) reflect.Type
```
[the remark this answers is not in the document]
vlastahajek: Honestly, I haven't noticed a difference before. The alias came from the original sketch and the second one was added later.
I will change that.

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+}
+
+// rest of supported type kinds
+const (
+	durationKind = fieldKind(255 + iota)
+	timeKind
+	bytesKind
+)
+
+// colType is integer based representation annotated CSV types
```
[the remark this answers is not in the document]
vlastahajek: Ok

annotatedcsv/decode.go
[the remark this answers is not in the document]
rogpeppe: tbh i'm not sure that `fieldTypeOf` is worth having - it's not used that much.

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+
+// fieldKindOf determines a fieldKind for given reflect.Type
+func fieldKindOf(t reflect.Type) fieldKind {
+	switch t {
+	case canonicalTypes[durationCol]:
+		return durationKind
+	case canonicalTypes[timeColRFC], canonicalTypes[timeColRFCNano]:
+		return timeKind
+	case canonicalTypes[base64BinaryCol]:
+		if t.Elem().Kind() == reflect.Uint8 {
```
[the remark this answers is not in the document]
vlastahajek: For example,  if `t` would be e.g. array of string.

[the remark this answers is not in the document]
rogpeppe: That can't happen because `canonicalTypes[base64BinaryCol]` is `[]byte` which can't match `[]string` (we're doing exact type matching here, not matching by kind).

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+	r.colSetters = setters
+	r.decodeType = t
+
+	return nil
+}
+
+// convertColumnValue set a value from current row of given column index
+// to a struct or a slice field value
+func (r *Reader) convertColumnValue(v reflect.Value, colIndex int) error {
+	s := stringTernary(r.row[colIndex], r.cols[colIndex].Default)
```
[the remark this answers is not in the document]
vlastahajek: Ok, it was used in more places, but after a refactor it is used only once. I will inline it

annotatedcsv/reader.go
vlastahajek: Good idea 👍

annotatedcsv/decode.go
vlastahajek: `flux` sounds, indeed, specific to the Flux results. As annotatedcsv is package for generally annotated CSV data, so `csv` is better.

annotatedcsv/decode.go
```diff
@@ -0,0 +1,451 @@
...
+// forEachField enumerates visible fields of t, finds field name and calls field function
+func forEachField(t reflect.Type, ff fieldFunc) error {
+	fields := ireflect.VisibleFields(t)
+	for _, f := range fields {
+		name := f.Name
+		if tag, ok := f.Tag.Lookup("flux"); ok {
+			if tag == "-" {
+				continue
+			}
+			name = tag
```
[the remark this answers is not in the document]
vlastahajek: Ok.

annotatedcsv/decode.go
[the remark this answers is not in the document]
vlastahajek: Ok, I see. So, sure,  this check is useless.

annotatedcsv/decode.go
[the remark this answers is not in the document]
vlastahajek: I'm refactoring the initColumns and Decode based on your comments, this is even not needed anymore.

annotatedcsv/decode.go
vlastahajek: I was trying this, but as `toInterface{}` needs `conversions`, it's a cycle dependency. So I've chosen `init` as a solution

annotatedcsv/decode_test.go
vlastahajek: In fact, this is the only unit test testing by private API. It is done this way to easily cover all type combinations, even those not possible using supported column types (mostly negative tests).
As others are testing using public API by inputting CSV data, I thought, this test could just use private API to be concise as possible

annotatedcsv/decode.go
vlastahajek: Ok

annotatedcsv/decode.go
vlastahajek: Ok

annotatedcsv/decode.go
rogpeppe: Good point! I hadn't realised that. Maybe add a comment noting that?

annotatedcsv/decode_test.go
rogpeppe: I'd still suggest doing this via the public API. It would be great to have as much freedom to refactor as possible. I think you should be able to get it just about as terse by using a bit of reflection.

annotatedcsv/decode.go
vlastahajek: Ok

annotatedcsv/decode.go
vlastahajek: Do you mean a comment which explains why it is initialized in the `init()`?

vlastahajek:
@rogger, I've pushed an update that addresses all remaining issues. Please, kindly check.

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+func (r *Reader) Decode(x interface{}) error {
+	t := reflect.TypeOf(x)
+	if err := r.initColumns(t, r.cols); err != nil {
+		return err
+	}
+	return r.decodeRow(reflect.ValueOf(x), r.row)
+}
+
+// initColumns initializes r.colSetters for the given columns.
+// It's called once per section.
```
rogpeppe, line 73: It's called every time `Decode` is called, not once per section.

```suggestion
// initColumns initializes r.colSetters for decoding the given columns
// into a value of the given type.
```

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+		fieldsMap := make(map[string]reflect.StructField)
+		fields := ireflect.VisibleFields(et)
+		for _, f := range fields {
+			name := f.Name
+			if tag, ok := f.Tag.Lookup("csv"); ok {
+				switch {
+				case tag == "-":
+					continue
+				case tag == "-,":
+					tag = "-"
```
rogpeppe, line 97: This case looks redundant to me - it's handled by the case below AFAICS.

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+				case strings.Contains(tag, ","):
```
rogpeppe, line 98: We can avoid the double search by doing:

```
if tag == "-" {
    continue
}
if i := strings.IndexByte(tag, ','); i >= 0 {
    if i != len(tag)-1 {
       return fmt.Errorf("tag attributes are not supported")
    }
    tag = tag[:i]
}
```

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+				}
+				name = tag
+			}
+			fieldsMap[name] = f
+		}
+
+		for i, col := range columns {
+			f, ok := fieldsMap[col.Name]
+			if !ok {
+				continue
```
rogpeppe, line 114:
```suggestion
				// The column isn't mentioned in the struct.
				continue
```

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+			}
+			ftype := f.Type
+			colType, ok := columnTypes[col.Type]
+			if !ok {
+				// ignore invalid type and use string
+				colType = stringCol
+			}
+			fkind := fieldKindOf(ftype)
+			convert, ok := conversions[conv{colType, fkind}]
```
rogpeppe, line 123:
```suggestion
			convert, ok := conversions[conv{colType, fieldKindOf(ftype)}]
```

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+			if !ok {
+				return fmt.Errorf("cannot convert from column type %s to %v", col.Type, ftype)
+			}
+			setters[i] = func(v reflect.Value, colIndex int) error {
+				if err := r.convertColumnValue(v.FieldByIndex(f.Index), colIndex, convert); err != nil {
+					return err
+				}
```
rogpeppe, line 130:
```suggestion
				return r.convertColumnValue(v.FieldByIndex(f.Index), colIndex, convert)
```

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+				return nil
+			}
+		}
+		r.decodeRow = func(v reflect.Value, row []string) error {
+			v = v.Elem()
```
rogpeppe, line 135:
```suggestion
			if v.IsNil() {
			    return fmt.Errorf("decode into nil %s", v.Type())
			}
			v = v.Elem()
```

or maybe a panic. Anyway, perhaps nicer to have an explicit message rather than a panic inside reflect.

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+				if err := c(v, i); err != nil {
+					return err
+				}
+			}
+			return nil
+		}
+	case reflect.Slice:
+		s := et.Elem()
+		if s != stringType && s != emptyInterfaceType {
+			return fmt.Errorf("cannot decode into pointer to slice of %v", s)
```
rogpeppe, line 149:
```suggestion
			return fmt.Errorf("cannot decode into *%v", s.Type())
```
?

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+			if e.Cap() < c {
+				e = reflect.MakeSlice(e.Type(), c, c)
+			} else {
+				e = e.Slice(0, c)
+			}
+			for i := 0; i < c; i++ {
+				if err := r.colSetters[i](e, i); err != nil {
+					return err
+				}
+			}
```
rogpeppe, line 182: Maybe slightly more efficient if the compiler isn't that clever:

```suggestion
			for _, setf := range r.colSetters {
			    if err := setf(e, i); err != nil {
			        return err
			    }
			}
```

annotatedcsv/decode.go
```diff
@@ -0,0 +1,463 @@
...
+const (
+	stringCol = colType(iota)
+	boolCol
+	durationCol
+	longCol
+	unsignedLongCol
+	doubleCol
+	base64BinaryCol
+	rfc3339Col
+	rfc3339NanoCol
```
rogpeppe, line 259: I don't think we need to distinguish between `rfc3339` and `rfc3339Nano` - I'd lose `rfc3330Nano`

inluxclient/query.go
```diff
@@ -98,12 +98,15 @@ func (r *QueryResultReader) errorSection() error {
...
-		message, ok1 := row[0].(string)
-		reference, ok2 := row[1].(int64)
-		if !ok1 || !ok2 {
+		if row[0] == "" {
+			return errors.New("no row found in error section")
+		}
+		var e QueryError
+		if err := r.Decode(&e); err != nil {
+
 			return fmt.Errorf("unexpected column types (%T, %T) in error section", row[0], row[1])
```
rogpeppe, line 107: I'd be inclined to keep the decode error in there.
```suggestion
			return fmt.Errorf("cannot decode row in error section: %v", err)
```

rogpeppe (approved):
This looks brilliant, thanks! A few more minor comments and suggestions, but LGTM.

annotatedcsv/decode.go
vlastahajek: Good point 👍

annotatedcsv/decode.go
vlastahajek: ok

PR #256, opened 2021-05-14
feat: allow Doer interface for http.Service

### LINKED ISSUE #255: Should allow passing in an interface to http.Service
__Proposal:__
Instead of requiring http.Service to have a concrete http.Client, I would like it to take a `Doer` interface, since it only actually uses the `Do` method.

__Current behavior:__
Currently it takes a concrete `http.Client`, which is more difficult to implement against.

__Desired behavior:__
I have filed a small PR to implement this, that's probably easier than describing in more detail.

__Alternatives considered:__
Copying the `http.Service` implementation and creating my own.

__Use case:__
I want to use this module in the `flux` cli, which only passes around and interface wrapping aspects of an `http.Client`, not a full client.

lesam:
Closes #255

## Proposed Changes

See issue #255

## Checklist

- [x] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+### Features
+[#256](https://github.com/influxdata/influxdb-client-go/pull/256) struct http.Service can be created with a 'Doer' interface
+
 ## 2.3.0 [2021-04-30]
 ### Breaking change
 [#253](https://github.com/influxdata/influxdb-client-go/pull/253) Interface 'Logger' extended with 'LogLevel() uint' getter.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## Unreleased
-
 ### Features
-[#256](https://github.com/influxdata/influxdb-client-go/pull/256) struct http.Service can be created with a 'Doer' interface
+[#256](https://github.com/influxdata/influxdb-client-go/pull/256) Allowing 'Doer' interface for HTTP requests
 
 ## 2.3.0 [2021-04-30]
 ### Breaking change
--- a/api/http/options.go
+++ b/api/http/options.go
@@ -82,14 +82,14 @@ type Doer interface {
 // Setting the HTTPDoer will cause the other HTTP options
 // to be ignored.
 func (o *Options) SetHTTPDoer(d Doer) *Options {
-	if d == nil {
-		return o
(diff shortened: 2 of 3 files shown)


lesam:
I need this to support this PR: https://github.com/influxdata/flux/pull/3718

vlastahajek (changes requested):
Thanks for the PR!
In general, supporting Doer interface is a good enhancement 👍 

Implementation can be less complex by adding `SetHTTPDoer` and `HTTPDoer()` do api/http/Options.

lesam:
I'm not sure that would be less complex.

We do need other methods (e.g. CloseIdleConnections) elsewhere, so we would have to error or panic when calling options.HTTPClient() if the client was not set up correctly - this seems worse than adding a constructor on Service.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Unreleased
+
+### Features
+[#256](https://github.com/influxdata/influxdb-client-go/pull/256) struct http.Service can be created with a 'Doer' interface
```
vlastahajek, line 4:
```suggestion
[#256](https://github.com/influxdata/influxdb-client-go/pull/256) Allowing 'Doer' interface  for HTTP requests
```

api/http/options.go
```diff
@@ -65,6 +71,32 @@ func (o *Options) OwnHTTPClient() bool {
...
+//
+// Setting the HTTPDoer will cause the other HTTP options
+// to be ignored.
+func (o *Options) SetHTTPDoer(d Doer) *Options {
+	if d == nil {
+		return o
+	}
+	o.doer = d
+	o.ownClient = false
+	return o
```
vlastahajek, line 90:
```suggestion
	if d != nil {
		o.doer = d
		o.ownClient = false
	}
	return o
```

api/http/options.go
```diff
@@ -65,6 +71,32 @@ func (o *Options) OwnHTTPClient() bool {
...
+}
+
```
vlastahajek, line 92:
```suggestion
// HTTPDoer returns actual Doer if set, or http.Client
```

vlastahajek (changes requested):
Thanks for the update. Just a couple of changes to-do yet.

vlastahajek (approved):

PR #267, opened 2021-08-03
fix: wrong content type header

### LINKED ISSUE #266: Wrong content type header
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

```go
bucketApi := client.BucketsAPI()
_, err = bucketApi.CreateBucketWithName(ctx, org, bucket, defaultBucketRetentionRule)
if err != nil {
    panic(err)
}
```
__Expected behavior:__
Create a bucket

__Actual behavior:__
panics with the following error: `invalid: invalid Content-Type header value`

__Specifications:__
 - Client Version: `github.com/influxdata/influxdb-client-go/v2 v2.4.0`
 - InfluxDB Version: cloud (btw, how can I get the exact version?)
 - Platform: debian (docker)
 
 __Possible cause:__

In [NewPatchBucketsIDRequestWithBody definition](https://github.com/influxdata/influxdb-client-go/blob/a842d5113a8780656bcee34a83e29c159e6faf04/domain/client.gen.go#L4292-L4337) the `content-type` header is added 2 times

FedeBev:
Closes #266 

## Proposed Changes

Updated openapi-codegen, it was quite old; this required a minor change to the client-with-response template that handles the status code > 299 (simply moved outside the switch).

The openapi-codegent client template contains a duplicated line `{{if .HasBody}}req.Header.Add("Content-Type", contentType){{end}}` leading to something like the following:
```go
func NewPatchBucketsIDRequestWithBody(server string, bucketID string, params *PatchBucketsIDParams, contentType string, body io.Reader) (*http.Request, error) {
...
	req.Header.Add("Content-Type", contentType)

	if params.ZapTraceSpan != nil {
...
	}

	req.Header.Add("Content-Type", contentType)
	return req, nil
}
```
The result is that the `content-type` header could look like this: `Content-Type: application/json; application/json` leading to the error reported in the issue (not alway though)

This PR resolves the issue.

## Checklist

- [ ] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x] Tests pass


vlastahajek:
@FedeBev, thanks for posting the issue and this fix. 
However, due to the [my 3 unmerged](https://github.com/deepmap/oapi-codegen/pulls/vlastahajek) PRs in the `oapi-codegen` we cannot use the original release. Yes, one of the PR could be added to the template, but the current implementation is more general. Some internal errors are returned as JSON, but there are not specified in the swagger.

PR #271, opened 2021-08-11
feat: use exponential retry strategy

vlastahajek:
## Proposed Changes

This PR aligns retry strategy implementations across all official InfluxDB client libraries so that a delay for the next retry delay is a random value in the interval _retryInterval * exponentialBase^(attempts)_ and _retryInterval * exponentialBase^(attempts+1)_.
The defaults were changed:
- retryInterval=5_000
- exponentialBase=2 (it is now also configurable)
- maxRetryDelay=125_000
- maxRetries=5

Retry delays are by default randomly distributed within the ranges of [5_000-10_000, 10_000-20_000, 20_000-40_000, 40_000-80_000, 80_000-125_000]. 

Added is also _MaxRetryTime_ option. When an overall time spent by retrying exceeds a maxRetryTime (180_000 millis by default), the write is not retried and fails.

## Checklist

- [X] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/api/write/options.go
+++ b/api/write/options.go
@@ -21,14 +21,18 @@ type Options struct {
 	useGZip bool
 	// Tags added to each point during writing. If a point already has a tag with the same key, it is left unchanged.
 	defaultTags map[string]string
-	// Default retry interval in ms, if not sent by server. Default 5,000ms
+	// Default retry interval in ms, if not sent by server. Default 5,000.
 	retryInterval uint
-	// Maximum count of retry attempts of failed writes
+	// Maximum count of retry attempts of failed writes, default 5.
 	maxRetries uint
-	// Maximum number of points to keep for retry. Should be multiple of BatchSize. Default 50,000
+	// Maximum number of points to keep for retry. Should be multiple of BatchSize. Default 50,000.
 	retryBufferLimit uint
-	// Maximum retry interval, default 5min (300,000ms)
+	// The maximum delay between each retry attempt in milliseconds, default 125,000.
 	maxRetryInterval uint
+	// The maximum total retry timeout in millisecond, default 180,000.
+	maxRetryTime uint
+	// The base for the exponential retry delay
+	exponentialBase uint
 }
 
 // BatchSize returns size of batch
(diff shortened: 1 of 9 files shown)


api/write/options.go
```diff
@@ -21,14 +21,18 @@ type Options struct {
...
-	// Maximum count of retry attempts of failed writes
+	// Maximum count of retry attempts of failed writes, default 5.
 	maxRetries uint
-	// Maximum number of points to keep for retry. Should be multiple of BatchSize. Default 50,000
+	// Maximum number of points to keep for retry. Should be multiple of BatchSize. Default 50,000.
 	retryBufferLimit uint
-	// Maximum retry interval, default 5min (300,000ms)
+	// The maximum delay between each retry attempt in milliseconds, default 125,000.
 	maxRetryInterval uint
+	// The maximum total retry timeout in millisecond, default 180,000.
```
sranka, line 32: I didn't find a test herein that proves that it is a maximum retry time since the first retry attempt ... is it really so?

vlastahajek: [This line](https://github.com/bonitoo-io/influxdb-client-go/blob/feat/rnd_exp_retry/internal/write/service_test.go#L395) is intended to test that.

sranka: This is IMHO testing `maxRetryInterval`, I was looking for a test (or proof) that `maxRetryTime` works.

sranka: The test is really difficult to understand as it is not obvious that a second `srv.HandleWrite` is required to return an error from the previously inserted expired batch ... a comment might have helped me to understand it. Anyway, It is not clear from the code that [the second WriteBatch](https://github.com/bonitoo-io/influxdb-client-go/blob/feat/rnd_exp_retry/internal/write/service_test.go#L273) is actually tried immediately, IMHO it is not. A highe knowldge of the context is required even for the test. A test that would match the user expectation is something that was looking for, simply checking that an error is signalized after maxRetryTime is reached and the retried item is removed from the retry queue.

vlastahajek: Yes, for someone unfamiliar with how the write service works this seems complicated. I added more comments to improved readability. 
One thing that is not obvious, and is that retries are not scheduled to be sent automatically. Retries are triggered by new writes.

sranka: I was also confused by the fact that `WriteBatch` can fail even without trying to write the data on input. It was this way before, so it is not part of this PR. Thank you for your explanation,

sranka (approved):

PR #272, opened 2021-08-11
expose the Doer to consumers/users

dovholuknf:
## Proposed Changes

these changes will allow a consumer to set the "Doer", allowing someone to establish the httpClient transport outside of the influx db client. the openziti project is a zero-trust sdk which allows for replacing the network connection with a more secure overlay network. To do that we need to be able to create "the socket/pipe" between the client and the server. Golang has wonderful abstractions for this already in `http.Client` (the Transport property). We are leveraging that property. I didn't find a clean way in the existing API to access/replace the httpClient's Transport but I did discover that api.http.service.go already had Doer exposed - so I figured that was the pattern that would be acceptable.

I'm open to changing it however you would like or need though?

## Checklist

- [ ] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x] A test has been added if appropriate (doesn't seem appropriate?)
- [?] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)
☝️ this returned an empty response - I would have signed it if I could have

### DIFF
--- a/client.go
+++ b/client.go
@@ -94,6 +94,13 @@ func NewClient(serverURL string, authToken string) Client {
 	return NewClientWithOptions(serverURL, authToken, DefaultOptions())
 }
 
+// NewClientWithDoer creates a Client for connecting to the InfluxDB server using the provided Doer interface.
+// authToken is an authentication token. It can be empty in case of connecting to newly installed InfluxDB server, which has not been set up yet.
+func NewClientWithDoer(doer http.Doer, serverURL, authToken string, options *Options) Client {
+	options.httpOptions.SetHTTPDoer(doer)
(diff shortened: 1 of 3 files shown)


vlastahajek:
@dovholuknf, `Doer` is already supported. Set it in the HTTPOptions to use it:
```go
doer := &http.Client{}
opts := influxdb2.DefaultOptions()
opts.HTTPOptions().SetHTTPDoer(doer)
client := influxdb2.NewClientWithOptions("http://localhost:8086", "my-token", opts);
```

dovholuknf:
sweet! I totally missed HTTPOptions. Thanks !

PR #276, opened 2021-09-08
fix: added mutex lock and unlock around read to logger struct for con…

shubhamcoc:
…current use

Closes #

## Proposed Changes

Mutex lock for reading to logger struct for concurrent clients. Race conditions occur if one client uses the Info function and the other client sets the log level.

## Checklist

- [x] Rebased/mergeable
- [x] Tests pass


shubhamcoc:
@vlastahajek kindly review this patch

PR #281, opened 2021-11-05
feat(v3): common HTTP handling code + query

vlastahajek:
This is a proposal of a general function that should help make other future functions, which will communicate with the server, easier. By future functions I mean those parts of the Client struct: Ready, Health, Delete, Write (also PointWriter).
Most reusable code is handling server errors. 

Retry-After reading is skipped here as it is needed to be handled only in the Write function to handle server back-pressure and thus Write can read this header itself. Or, it can be added later.

Added is also the `Query` function as the first feature using the `makeAPICall` function.


inluxclient/client.go
```diff
@@ -71,3 +91,83 @@ func New(params Params) (*Client, error) {
...
+}
+
+// resolveHTTPError parses server error response and returns error with human readable message
+func (c *Client) resolveHTTPError(r *http.Response) *ServerError {
+	// successful status code range
+	if r.StatusCode >= 200 && r.StatusCode < 300 {
+		return nil
+	}
+
+	httpError := &ServerError{
```
rogpeppe, line 135: I think we could probably try both error kinds here with one decode, which could simplify the logic quite a bit:

```
var httpError struct {
     ServerError
     // This is used by the InfluxDB v1 responses.
     Error string `json:"error"`
}
httpError.StatusCode = r.StatusCode
... decode into httpError
if httpError.Code == "" && httpError.Message == "" {
    httpError.Message = httpError.Error
}

```

inluxclient/client.go
```diff
@@ -71,3 +91,83 @@ func New(params Params) (*Client, error) {
...
+				httpError.Message = fmt.Sprintf("cannot decode error response: %v", err)
+			}
+			if httpError2.Error != "" {
+				httpError.Message = httpError2.Error
+			}
+		}
+	}
+	if httpError.Message == "" {
+		if len(body) > 0 {
+			httpError.Message = string(body)
```
rogpeppe, line 166: Unfortunately in some cases, this could be a very large piece of HTML produced by some proxy. In the past, I've resorted to trying to get some readable text out of the HTML, but for now maybe just add a TODO: "This could be a large piece of unreadable body; we might be able to do better than this" ?

inluxclient/client_query.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// QueryParam is key-value pair for parameters inside the Flux query
+type QueryParam struct {
+	Key   string
+	Value string
+}
+
+// Query defines model for Query.
+type Query struct {
```
rogpeppe, line 28: "model" doesn't really mean that to me FWIW, and this struct isn't part of the public API, so how about this?

```suggestion
// queryBody holds the body for an HTTP query request.
type queryBody struct {
```

inluxclient/client_query.go
```diff
@@ -0,0 +1,68 @@
...
+var defaultDialect = Dialect{
+	Annotations: []string{"datatype", "default", "group"},
+	Delimiter:   ",",
+	Header:      true,
+}
+
+// Query sends the given Flux query on the given organization ID.
+// The result must be closed after use.
+// Flux query can contain reference to params that must be passed in the params argument
+func (c *Client) Query(ctx context.Context, org, query string, params ...QueryParam) (*QueryResultReader, error) {
```
rogpeppe, line 45: I'm wondering if there's a good use case for having a client talking to many different organization IDs. ISTM that almost every realistic use case would involve the client talking to a single organization, so perhaps we could move the org ID to the client parameters, meaning that we wouldn't need to have it in every API endpoint? WDYT?

Another possible thought: instead of defining a `QueryParam` struct, we could accept a slice of alternating key-value strings, like strings.NewReplacer](https://pkg.go.dev/strings#NewReplacer) does, or maybe just a `map[string]string` ?

inluxclient/client_query.go
```diff
@@ -0,0 +1,68 @@
...
+	queryURL, _ := c.apiURL.Parse("query")
+
+	q := Query{Dialect: defaultDialect, Query: query, Type: "flux", Params: make(map[string]string)}
```
rogpeppe, line 48: Arguably clearer with the fields on separate lines.

inluxclient/client_query.go
```diff
@@ -0,0 +1,68 @@
...
+	for _, r := range params {
+		q.Params[r.Key] = r.Value
+	}
+	qrJSON, err := json.Marshal(q)
+	if err != nil {
+		return nil, err
+	}
+	resp, herr := c.makeAPICall(ctx, httpParams{
```
rogpeppe, line 56: No need to avoid reusing `err` AFAICS.

```suggestion
	resp, err := c.makeAPICall(ctx, httpParams{
```

inluxclient/client.go
```diff
@@ -71,3 +91,83 @@ func New(params Params) (*Client, error) {
 	}
 	return c, nil
 }
+
+// makeAPICall issues an HTTP request to InfluxDB server API url according to parameters.
+// Additionally, sets Authorization header and User-Agent.
+// It returns http.response or Error
+func (c *Client) makeAPICall(ctx context.Context, params httpParams) (*http.Response, *ServerError) {
```
rogpeppe, line 98: I think I'd suggest returning `error` rather than `*ServerError`. Returning a non-interface type as an error runs a distinct risk of [nil-vs-non-nil error confusion](https://go.dev/doc/faq#nil_error), and I think it's nice for the caller to be able to distinguish between a local error (as returned in the error from `http.Client.Do`) and a server-returned error. We can document that the error will always be a `*ServerError` when we got a response from the server).

inluxclient/client_query_test.go
```diff
@@ -0,0 +1,54 @@
...
+		w.WriteHeader(200)
+		w.Write([]byte(`#group,false,false,true,true,false,true,true,true,false,false,false
+#default,_result,,,,,,,,,,
+#datatype,string,long,dateTime:RFC3339,dateTime:RFC3339,dateTime:RFC3339,string,string,string,double,double,double
+,result,table,_start,_stop,_time,deviceId,location,sensor,air_hum,air_press,air_temp
+,,0,2021-10-19T14:39:57.464357168Z,2021-10-19T14:54:57.464357168Z,2021-10-19T14:40:21.833564544Z,2663346492,saman-home-room-0-1,BME280,48.8,1022.28,22.73
+,,0,2021-10-19T14:39:57.464357168Z,2021-10-19T14:54:57.464357168Z,2021-10-19T14:41:29.840881203Z,2663346492,saman-home-room-0-1,BME280,49.2,1022.34,22.7`))
+	}))
+	defer ts.Close()
+	client, err := New(Params{ServerURL: ts.URL, AuthToken: ""})
```
rogpeppe, line 25:
```suggestion
	client, err := New(Params{ServerURL: ts.URL})
```

rogpeppe (commented):
This is going in a nice direction, thanks! I've got a few comments and suggestions, but it's looking good in general.

inluxclient/version.go
```diff
@@ -0,0 +1,15 @@
...
+// license that can be found in the LICENSE file.
+
+package influxclient
+
+import (
+	"runtime"
+)
+
+// Version defines current version
+const Version = "3.0.0alpha1"
```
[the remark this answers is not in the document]
vlastahajek: Ok

inluxclient/client.go
vlastahajek: Ok

inluxclient/client.go
vlastahajek: Ok

inluxclient/client_query.go
vlastahajek: Ok

inluxclient/client_query.go
vlastahajek: The slice of alternating key-value slice strings is a very nice new pattern for me 👍. I like this more than the map because it allows using variadic function.

inluxclient/client_query.go
vlastahajek: Ok

inluxclient/client_query.go
vlastahajek: It was there to distinguish between error and ServerError. If `makeAPICall` returns `error`, it can be removed.

inluxclient/client.go
vlastahajek: Ok

inluxclient/client_query.go
vlastahajek: Regarding OrgID, interesting idea. If we have OrgID as top-level argument it will simplify many future API functions.
Indeed, most users use the client for a single organization. I've discussed this with colleagues, authors of the other official clients. One of them mentioned a customer that  uses C# client to connect to multiple organizations, but it seems it is a minor use-case.

rogpeppe: Yeah, I think if someone wants to talk to multiple orgs, it wouldn't be too much hassle to have multiple clients, or create a client for each request.

inluxclient/client.go
```diff
@@ -71,3 +93,77 @@ func New(params Params) (*Client, error) {
 	}
 	return c, nil
 }
+
+// makeAPICall issues an HTTP request to InfluxDB server API url according to parameters.
+// Additionally, sets Authorization header and User-Agent.
```
sranka, line 98: Shouldn't we also set `Content-Type` to `application/json` by default when body is non-nil

sranka (changes requested):
LGTM expect a few things
- `inluxclient` directory should be renamed
- query example should be added
- I didn't find any concept that would let the user get column values according to a returned column type

inluxclient/client.go
```diff
@@ -6,8 +6,13 @@
 package influxclient
```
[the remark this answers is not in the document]
vlastahajek: Good catch, thanks!

inluxclient/client.go
vlastahajek: Future `WritePoints` could use this function and it will set a different content-type

sranka: OK, setting the content type is always the responsibility of the caller.

sranka (changes requested):
LGTM, I only think that we might provide a better query example - simplify query parameters and the executed flux query.

influxclient/examples_test.go
```diff
@@ -0,0 +1,68 @@
...
+		23.0,
+	}
+
+	query := `from(bucket: params.from) 
+		|> range(start: duration(v: params.start)) 
+		|> filter(fn: (r) => r._measurement == "environment")
+		|> filter(fn: (r) => r._field == params.field)
+		|> filter(fn: (r) => r._value > params.value)`
+
+	res, err := client.Query(context.Background(), query, cond)
```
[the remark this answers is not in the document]
vlastahajek: Ok

sranka (approved):
LGTM :+1:

PR #287, opened 2021-11-25
feat: adding FluxRecord.Result() function

### LINKED ISSUE #283: Query API does not expose table's name
I have a Flux query that yields two distinct tables:

```
import "generate"

generate.from(count: 1, fn: (n) => 0, start: now(), stop: now())
  |> yield(name: "foo")

generate.from(count: 1, fn: (n) => 1, start: now(), stop: now())
  |> yield(name: "bar")
```

The resulting CSV is formatted like this:

```csv
#group,false,false,false,false
#datatype,string,long,dateTime:RFC3339,long
#default,foo,,,
,result,table,_time,_value
,,0,2021-11-09T20:50:58.440510582Z,0

#group,false,false,false,false
#datatype,string,long,dateTime:RFC3339,long
#default,bar,,,
,result,table,_time,_value
,,0,2021-11-09T20:50:58.440510582Z,1
```

The raw data view in the UI understands the table names:

Screen Shot 2021-11-09 at 3 52 59 PM

However, the Go client only exposes the table numbers: https://pkg.go.dev/github.com/influxdata/influxdb-client-go/v2@v2.5.1/api/query#FluxTableMetadata

AFAICS there is no way to access the table name, which prevents me from easily consuming queries with yield. To the best of my knowledge, with real queries that take time to execute, the table ID is unpredictable and cannot be used to determine which table is which.

vlastahajek:
Closes #283

## Proposed Changes

Added *FluxRecord.Result()* function as a convenient way to retrieve the Flux result name of data.
The `Result()` function extends collection of already existing accessor functions: `Table(), Start(), Stop(), Time(),  Measurement() and Value()`

## Checklist

- [X] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 ### Features
 - [#285](https://github.com/influxdata/influxdb-client-go/pull/285) Added *Client.Ping()* function as the only validation method available in both OSS and Cloud.
+- [#287](https://github.com/influxdata/influxdb-client-go/pull/287) Added *FluxRecord.Result()* function as a convenient way to retrieve the Flux result name of data.
 
 ### Bug fixes
 - [#285](https://github.com/influxdata/influxdb-client-go/pull/285) Functions *Client.Health()* and *Client.Ready()* correctly report an error when called against InfluxDB Cloud.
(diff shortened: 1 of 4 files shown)


bednar (approved):
LGTM 🍺

PR #294, opened 2021-12-22
fix: retain full error information for blocking writes (#293)

### LINKED ISSUE #293: WriteAPIBlocking discards http2.Error content
The fixes for issue #235 included: https://github.com/influxdata/influxdb-client-go/blob/4218b5069e003d9575d6d8221a43f7eed6e5f6ec/api/writeAPIBlocking.go#L76-L78 which unwraps the error returned from the lower levels.

When the error is an [http2.Error](https://pkg.go.dev/github.com/influxdata/influxdb-client-go/v2@v2.6.0/api/http#Error), e.g. because the token is unauthorized, this invokes: https://github.com/influxdata/influxdb-client-go/blob/4218b5069e003d9575d6d8221a43f7eed6e5f6ec/api/http/error.go#L34-L36 which completely discards all the Influx-specific information, producing an error of type `errors.errorString` that has nothing but the text.

So, unlike the non-blocking API, an error from WriteAPIBlocking() does *not* return an `http2.Error`, making it complex to understand why the write failed.  The idiom proposed in https://github.com/influxdata/influxdb-client-go/issues/292#issuecomment-999762597 cannot recover the lost information.

pabigot:
Closes #293

## Proposed Changes

## Checklist

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/api/writeAPIBlocking.go
+++ b/api/writeAPIBlocking.go
@@ -74,7 +74,7 @@ func NewWriteAPIBlocking(org string, bucket string, service http2.Service, write
 func (w *writeAPIBlocking) write(ctx context.Context, line string) error {
 	err := w.service.WriteBatch(ctx, iwrite.NewBatch(line, w.writeOptions.RetryInterval(), w.writeOptions.MaxRetryTime()))
 	if err != nil {
-		return err.Unwrap()
+		return err
 	}
 	return nil
 }


pabigot:
I should point out that the tests do *not* pass with this; I get:
```
tirzah[326]$ pwd
/mnt/devel/external/Tick/influxdb-client-go
tirzah[327]$ go test
--- FAIL: TestUserAgent (0.20s)
    client_test.go:95: err <nil>
    client_test.go:96: 
        	Error Trace:	client_test.go:96
        	Error:      	Received unexpected error:
        	            	<nil>
        	Test:       	TestUserAgent
--- FAIL: TestServerOnPath (0.00s)
    client_test.go:128: 
        	Error Trace:	client_test.go:128
        	Error:      	Received unexpected error:
        	            	<nil>
        	Test:       	TestServerOnPath
FAIL
exit status 1
FAIL	github.com/influxdata/influxdb-client-go/v2	3.514s
```

However I also get that on current master, so it's not due to this change.

pabigot:
Note that the solution is slightly complex because in some cases a nil `*http2.Error` is returned by the underlying function, which results in an `error` that has a concrete type with a nil value, which is not itself nil.

api/writeAPIBlocking.go
```diff
@@ -74,7 +74,7 @@ func NewWriteAPIBlocking(org string, bucket string, service http2.Service, write
 func (w *writeAPIBlocking) write(ctx context.Context, line string) error {
 	err := w.service.WriteBatch(ctx, iwrite.NewBatch(line, w.writeOptions.RetryInterval(), w.writeOptions.MaxRetryTime()))
 	if err != nil {
-		return err.Unwrap()
+		return err
 	}
 	return nil
```
vlastahajek, line 79: Why not then just:
```go
return w.service.WriteBatch(ctx, iwrite.NewBatch(line, w.writeOptions.RetryInterval(), w.writeOptions.MaxRetryTime()))
```

pabigot: Because the return value from that is type `*http2.Error`.  If its value is `nil`, then converting it to the `error` interface produces a value that has a bound type of `*http2.Error`, but nil pointer.  That interface value does not compare equal to `nil` in the caller, making it appear an error occurred when it didn't.

(Also, because I tried that, and the tests failed due to the above problem.)

vlastahajek: Yes, yes, I realized that at the same time as you replied. Sorry.

PR #296, opened 2022-01-01
feat: add blocking API to write a batch of records

### LINKED ISSUE #295: Add WriteBatch to WriteAPIBlocking
__Proposal:__
Extend the blocking API to allow writing a batch of already-aggregated line protocol records.

__Current behavior:__
The blocking API only accepts individual points, or points already converted to records.  The implemented functions then go on to do the conversion and aggregation to create a batch.

__Desired behavior:__
To write a batch that I've created myself.

__Use case:__
I'm continuing my attempt to use this package in a context where I can't tolerate lost data due to the server being inaccessible.

The non-blocking API has a variety of problems raised in other issues, but the showstopper is that it will drop data without allowing the application to recover the data; e.g. in [this path](https://github.com/influxdata/influxdb-client-go/blob/4218b5069e003d9575d6d8221a43f7eed6e5f6ec/internal/write/service.go#L142) but also in a [shutdown situation](https://github.com/influxdata/influxdb-client-go/pull/291#issuecomment-993673829).

The documentation of [WriteAPIBlocking](https://pkg.go.dev/github.com/influxdata/influxdb-client-go/v2@v2.6.0/api#WriteAPIBlocking) notes that it should be used when:

> there is a need to have explicit control of failed batches.

The problem in my situation is that `WriteAPIBlocking` accepts only on individual points or line-protocol records (possibly in a slice, but still separated).  In my application, if I need to persist unwritten points the natural encoding is newline-separated (or terminated) line protocol records in a byte sequence.  However, when I read that back off disk and prepare to resend it the current API requires me to split it into individual records at the newline, so that WriteRecord can then add the newlines back and aggregate it again.

This is cumbersome, but trivially avoided by extending the API to allow writing a batch.

pabigot:
Closes #295

## Proposed Changes

This extends the blocking API to allow writing line protocol records that have already been combined into a string with newline separators.

## Checklist

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/api/writeAPIBlocking.go
+++ b/api/writeAPIBlocking.go
@@ -58,6 +58,8 @@ type WriteAPIBlocking interface {
 	// WritePoint writes without implicit batching. Batch is created from given number of points
 	// Non-blocking alternative is available in the WriteAPI interface
 	WritePoint(ctx context.Context, point ...*write.Point) error
+	// WriteBatch writes line protocol record(s) that have already been aggregated into a string with newline separators into bucket.
+	WriteBatch(ctx context.Context, batch string) error
 }
 
 // writeAPIBlocking implements WriteAPIBlocking interface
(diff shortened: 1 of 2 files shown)


vlastahajek:
Thanks for the PR.  The new function to write a single string is redundant. The `WriteRecord` function can also take just a single string.
It makes sense to improve this function to skip the loop and also to skip creating a string buffer if there is just a single line.

pabigot:
OK, I can create a new PR that implements that behavior change, but it's more complicated than it first sounds, and possibly introduces a performance hit by having to check for the conditions under which newlines must be inserted/appended.  New API really seems more efficient and less confusing.

vlastahajek:
A single line doesn't need to have new-line-char. For such a simple case, breaking the API with a new feature seems to be too much.

PR #297, opened 2022-01-03
chore: optimize single-record blocking write

pabigot:
## Proposed Changes

This optimizes `WriteAPIBlocking.WriteRecord` in the case where a single string is provided, and explicitly documents that batches (records separated/terminated by newline) can be passed.  A test is added to confirm that this works as expected.

Unless the server chokes on empty lines this should resolve the need that motivated #295.

## Checklist

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/api/writeAPIBlocking.go
+++ b/api/writeAPIBlocking.go
@@ -51,7 +51,8 @@ import (
 //	}
 type WriteAPIBlocking interface {
 	// WriteRecord writes line protocol record(s) into bucket.
-	// WriteRecord writes without implicit batching. Batch is created from given number of records
+	// WriteRecord writes without implicit batching. Batch is created from given number of records.
+	// Individual arguments can also be batches (multiple records separated by newline).
 	// Non-blocking alternative is available in the WriteAPI interface
 	WriteRecord(ctx context.Context, line ...string) error
 	// WritePoint data point into bucket.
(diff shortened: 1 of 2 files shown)


api/writeAPIBlocking.go
```diff
@@ -80,18 +81,25 @@ func (w *writeAPIBlocking) write(ctx context.Context, line string) error {
...
-		for _, line := range line {
-			b := []byte(line)
-			b = append(b, 0xa)
-			if _, err := sb.Write(b); err != nil {
-				return err
-			}
+	if len(line) == 0 {
+		return nil
+	}
+	if len(line) == 1 {
```
sranka, line 87: empty lines are ignored by InfluxDB, adding extra new line is not required, it could simply `return w.write(ctx, line[0])` 

https://github.com/influxdata/influxdb/blob/11c00813f106062829759a97673c7c945d5859ca/models/points.go#L369

api/writeAPIBlocking.go
```diff
@@ -80,18 +81,25 @@ func (w *writeAPIBlocking) write(ctx context.Context, line string) error {
...
+		ln := line[0]
+		if ln[len(ln)-1] != '\n' {
+			ln += "\n"
 		}
-		return w.write(ctx, sb.String())
+		return w.write(ctx, ln)
 	}
-	return nil
+	var sb strings.Builder
```
sranka, line 94: a nit: this could be also optimized to avoid useless allocations of temporary buffers, it could Grow in advance to `len(line)+sum_of_line_lengths`

pabigot: `strings.Join(lines, "\n")` would probably be best, except for the need to add a terminal newline.  Lifting `[]bytes` out of the existing loop and expanding it at need based on line length could also work.  I didn't want to make that large a change, but if it's important enough I'll take a stab at it.

api/writeAPIBlocking.go
pabigot: The existing implementation assumes there are no terminating newlines on the passed records; it's necessary to add one when missing to keep the existing behavior.  Also, not adding a newline causes the tests to fail, and keeps this consistent with what's produced by line-protocol, see influxdata/line-protocol#52.

pabigot: BTW: That existing behavior is verified by the tests, which is why it isn't the way you suggested....

api/writeAPIBlocking.go
sranka: yes `strings.Join` is already an optimized code and it would be the best option, though it would probably break the tests. The original goal herein (doc + simple optimization for 1 line) is already the necessary improvement.

The level and depth of optimizations is up to you ... it was just a nit from me. @vlastahajek is the approver herein.

vlastahajek (approved):
@pabigot, Thanks for the improvement and the PR!

PR #308, opened 2022-02-17
feat: adding flux query parameters

### LINKED ISSUE #146: Parametrically build query, rather than creating large string
Currently, influxdb-client-go (and other InfluxDB 2.0 client libraries) uses a string to query an InfluxDB database. This works well if the query is static, but if the query is dynamic, it can lead to some problems if IDs of measurements/fields/names or values of time have to be dynamically entered into the query. Additionally, if a user was given control over what some measurements/fields/names are called at runtime, it could lead to potential SQL injection. 

Ideally, it should be possible to parametrically build a query out of variables or values; for example, calling a func/type to create a name or filter.

vlastahajek:
## Proposed Changes

Added support for [Flux query parameters](https://docs.influxdata.com/influxdb/cloud/query-data/parameterized-queries/), available in InfluxDB Cloud only at this moment.

`QueryAPI` is extended with `QueryWithParams` and `QueryRawWithParams`.

Query parameters can be passed as a struct or map. Param values can be only simple types or `time.Time`.
The name of a struct field or a map key (must be a string) will become a param name
The name of the parameter represented by a struct field can be specified by JSON annotation:

```go
type Condition struct {
     Start  time.Time  `json:"start"`
     Field  string     `json:"field"`
     Value  float64    `json:"value"`
}
```

Additionally, this requires changing the minimum Go version to 1.17.

## Checklist

- [x] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x] Tests pass

Closes #146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - [#304](https://github.com/influxdata/influxdb-client-go/pull/304) Added public constructor for `QueryTableResult`
 - [#307](https://github.com/influxdata/influxdb-client-go/pull/307) Synced generated server API with the latest [oss.yml](https://github.com/influxdata/openapi/blob/master/contracts/oss.yml). 
 - [#308](https://github.com/influxdata/influxdb-client-go/pull/308) Added Flux query parameters. Supported by InfluxDB Cloud only now.
+- [#308](https://github.com/influxdata/influxdb-client-go/pull/308) Go 1.17 is required
 
 ## 2.7.0[2022-01-20]
 ### Features 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ### Features
 - [#304](https://github.com/influxdata/influxdb-client-go/pull/304) Added public constructor for `QueryTableResult`
 - [#307](https://github.com/influxdata/influxdb-client-go/pull/307) Synced generated server API with the latest [oss.yml](https://github.com/influxdata/openapi/blob/master/contracts/oss.yml). 
+- [#308](https://github.com/influxdata/influxdb-client-go/pull/308) Added Flux query parameters. Supported by InfluxDB Cloud only now.
 
 ## 2.7.0[2022-01-20]
 ### Features 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ jobs:
             mkdir -p /tmp/artifacts
             mkdir -p /tmp/test-results
       - run: sudo rm -rf /usr/local/go
-      - run: wget https://golang.org/dl/go1.13.14.linux-amd64.tar.gz -O /tmp/go.tgz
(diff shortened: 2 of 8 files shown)


bednar (commented):
Awesome improvement 👍

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 ### Features
 - [#304](https://github.com/influxdata/influxdb-client-go/pull/304) Added public constructor for `QueryTableResult`
 - [#307](https://github.com/influxdata/influxdb-client-go/pull/307) Synced generated server API with the latest [oss.yml](https://github.com/influxdata/openapi/blob/master/contracts/oss.yml). 
+- [#308](https://github.com/influxdata/influxdb-client-go/pull/308) Added Flux query parameters. Supported by InfluxDB Cloud only now.
+- [#308](https://github.com/influxdata/influxdb-client-go/pull/308) Go 1.17 is required
```
sranka, line 6: Does it depend on 1.17 features? Isn't it the case that 1.17 is now used by tests, so it is not a feature.

sranka (changes requested):

CHANGELOG.md
sranka: It should IMHO at least follow [Go Security Policy](https://go.dev/security) so that each major Go release is supported until there are two newer major releases.

sranka: Thank you for the explanation. 1.18 is likely to come this month and 1.16 will thus go out of support.  So it makes sense to require 1.17 herein.

sranka (approved):

deefdragon:
Would it be worth making a PR to update the docs to warn that this is not supported in OSS (yet)? Its kinda strange its in the readme but not the docs.

vlastahajek:
There is a note in the Readme about this: https://github.com/influxdata/influxdb-client-go#parametrized-queries

deefdragon:
I acknowledge the note in the Readme. My question is if it should be added to the interface doc comments.

I didn't use the Readme to figure out how to use the library. I used the docs for the interface itself on pkg.go.dev, and had no idea what error I was getting until I got lucky with a Google search showing the ⚠️.

vlastahajek:
I understand your point. 
The problem with such a note in the API doc is that when support in OSS is ready, the note must be deleted and a new version of the library must be released. even if there is no change in the code,

deefdragon:
The same applies to the Readme? And even if OSS gets support soon, some people might not update their dbs for a long while, meaning the warning will be relevent for some Time after. 
Regardless, I'm not saying to add "this won't work" I'm saying to add a warning that it _might not_ work based on the version. Or atleast tell the user that if they encounter a params unknown error what to do.

> Support for parameterized queries is database version dependant. If you encounter a keyword params unknown error, check your influxdb version.

vlastahajek:
Readme is more often read on GitHub than on the Go API docs site and can be updated on the fly, I think.
However, your suggestions make sense. I will be updated.

ylhan:
When will support for parametrized queries be added to OSS? What's the justification for not including it?

PR #313, opened 2022-03-04
feat: add User-Agent header overwrite possibility

yohan-couppey:
This is a proposal to add the possibility to overwrite the User-Agent field when we do requests.

Example of use : 

```
	c := client.NewClient( fmt.Sprintf("https://%s:%s", "influx_url", "influx_proxy_port"), "influx_token" )
	defer c.Close()

	c.HTTPService().SetUserAgent("custom_user_agent")
```


yohan-couppey:
The whole code was not committed, I am sorry for the inconvenience and I will make a new pull request when it's done

PR #314, opened 2022-03-04
feat: add User-Agent header overwrite possibility

yohan-couppey:
## Proposed Changes

This is a proposal to add the possibility to overwrite the User-Agent field when we do requests.

Example of use :

```
c := client.NewClient( fmt.Sprintf("https://%s:%s", "influx_url", "influx_proxy_port"), "influx_token" )
defer c.Close()

c.HTTPService().SetUserAgent("custom_user_agent")
```

## Checklist

- [ ] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x]  A test has been added if appropriate (doesn't seem appropriate)
- [ ] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/api/http/service.go
+++ b/api/http/service.go
@@ -44,6 +44,10 @@ type Service interface {
 	SetAuthorization(authorization string)
 	// Authorization returns current authorization header value
 	Authorization() string
+	// SetUserAgent sets the User-Agent header value
+	SetUserAgent(authorization string)
+	// UserAgent returns current User-Agent header value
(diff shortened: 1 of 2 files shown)


yohan-couppey:
I agree that we can do it with the Doer interface, but it was more convenient as User-Agent is as common as the Authorization field and avoid a lot of code. 

But I will continue to do it with the Doer interface then.

Issue #317, opened 2022-03-21
Panic: close on closed channel when multiple goroutines call `influxdb2.Client.Close()`


dleonard203:
It is possible to get the `influxdb2.Client` to panic if `Client.Close()` is called at the same time from multiple goroutines:

```
panic: close of closed channel
goroutine 341 [running]:
github.com/influxdata/influxdb-client-go/v2/api.(*WriteAPIImpl).Close(0xc0001b15e0)
        /go/pkg/mod/github.com/influxdata/influxdb-client-go/v2@v2.2.0/api/write.go:176 +0x6f
github.com/influxdata/influxdb-client-go/v2.(*clientImpl).Close(0xc00064e0c0)
        /go/pkg/mod/github.com/influxdata/influxdb-client-go/v2@v2.2.0/client.go:218 +0x8e
```

It appears the cause is that on calling `Client.Close`, it is [looping through all](https://github.com/influxdata/influxdb-client-go/blob/v2.8.1/client.go#L246) non-blocking WriteAPI's, and [closing a channel on the struct](https://github.com/influxdata/influxdb-client-go/blob/v2.2.0/api/write.go#L176). Before doing this channel closing, the `Close` method mkes sure the channel is not nil. The channel is set to `nil` _after_ some processing is done, so two goroutines can pass this `nil` check, causing them both to attempt to close the channel.

I think a fix to this would be to place a `shutdown` mutex on the non-blocking WriteAPI object, lock it when `WriteAPI.Close()` is called, set the channel to `nil` with the lock, and check for the channel being `nil` after the lock is acquired. That way, if two goroutines call `client.Close()` at the same time, the second goroutine will see that the first has set the channel to `nil`, and will not attempt the shutdown processing.

__Steps to reproduce:__
1. Create one `influxdb2.Client` object to be accessed from multiple goroutines
2. Start a goroutine reading from influxdb every second, calling a `defer client.Close()`
3. Start a goroutine writing 300 points once per second using non-blocking WriteAPI, also calling `defer client.Close()` at the end of its call
4. There should be a panic within a few seconds

__Expected behavior:__
There should be no panic if two goroutines try to close the client.

__Actual behavior:__
The client panics

__Specifications:__
 - Client Version: 2.2.0
 - InfluxDB Version: 2.1
 - Platform: Linux

Edit: I haven't used the latest client, but checked the 2.8.1 tag and see the same issue should still exist in latest.

vlastahajek:
This was fixed by #319

PR #319, opened 2022-03-26
fix: add concurrency panic protection around Client and WriteAPIImpl Close

dleonard203:
## Proposed Changes

This PR aims to address https://github.com/influxdata/influxdb-client-go/issues/317 . If multiple goroutines call `influxdb2.Client.Close()`, or `api.WriteAPI.Close()`, it is possible for there to be a panic due to closing a closed channel.

## Checklist

- [X] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass
- [X] Sign [CLA](https://www.influxdata.com/legal/cla/) (if not already signed)

### DIFF
--- a/api/write.go
+++ b/api/write.go
@@ -7,6 +7,7 @@ package api
 import (
 	"context"
 	"strings"
+	"sync"
 	"time"
 
 	http2 "github.com/influxdata/influxdb-client-go/v2/api/http"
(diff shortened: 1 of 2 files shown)


api/write.go
```diff
@@ -194,6 +197,8 @@ x:
 // Close finishes outstanding write operations,
 // stop background routines and closes all channels
 func (w *WriteAPIImpl) Close() {
+	w.closingMu.Lock()
```
dleonard203, line 200: Since this is an exported type, users could (and may already) call `WriteAPIImpl{}` without instantiating this new mutex, so it could default to `nil`. Does doing a `nil` check make sense here before acquiring the lock?

vlastahajek: As there is no nil check for other fields, it is not necessary.

vlastahajek (approved):
Thanks for the issue and the fix!

PR #341, opened 2022-06-30
fix: logging messages about discarding batch as error

### LINKED ISSUE #340: Flush doesn't perform retries of write failures
The default write retry logic is documented to retry 5 times before failing.

It seems this may not in fact be the case.

__Steps to reproduce:__
List the minimal actions needed to reproduce the behaviour.

I have created a test showing what I believe to be the broken behaviour: https://github.com/influxdata/influxdb-client-go/pull/339

__Expected behavior:__

`Flush` should wait for all retries to complete

__Actual behavior:__

It seems `Flush` does not wait for retries and the data appears to be lost.

__Specifications:__
 - Client Version: current

vlastahajek:
## Proposed Changes

Changing logging level of messages about discarding batch to Error. A part of the fix for #340

## Checklist

- [X] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [unreleased]
+### Bug fixes
+- [#341](https://github.com/influxdata/influxdb-client-go/issues/341) Changing logging level of messages about discarding batch to Error. 
+
 ## 2.9.1 [2022-06-24]
 ### Bug fixes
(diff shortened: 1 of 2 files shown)


sranka (approved):
LGTM

PR #354, opened 2022-09-09
fix: efficient write blocking synchronization

### LINKED ISSUE #352: Improve throughput of blocking writes to the same bucket
__Proposal:__

High-level: The throughput of blocking writes to the same bucket is limited by a broad-scoped mutex.

Lower-level: Remove the mutex from `WriteAPIBlocking` implementation entirely, or at least reduce its scope to only exactly where it's needed.

__Current behavior:__

The method `writeAPIBlocking.write` locks a mutex at the start and doesn't unlock it until done (see [source](https://github.com/influxdata/influxdb-client-go/blob/cffdd2bde634fef240165a37750952aaf551290c/api/writeAPIBlocking.go#L80)). The mutex seems necessary to guard `w.batching` and `w.batch` but not `w.service` nor `w.writeOptions` which can't be changed after creation.

__Desired behavior:__

1. Remove the mutex entirely: No easy way to do this without breaking API changes
2. Reduce the scope of the mutex: Since the mutex is only necessary to guard implicit-batching&ndash;related parameters, it does not need to be locked for the entire method call.

Here's a possible solution for option 2:

```go
func (w *writeAPIBlocking) write(ctx context.Context, line string) error {
	body := line
	func(){
		w.mu.Lock()
		defer w.mu.Unlock()
		if w.batching {
			w.batch = append(w.batch, line)
			if len(w.batch) == int(w.writeOptions.BatchSize()) {
				body = strings.Join(w.batch, "\n")
				w.batch = w.batch[:0]
			} else {
				body = ""
			}
		}
	}()
	if body == "" {
		return nil
	}
	err := w.service.WriteBatch(ctx, iwrite.NewBatch(body, w.writeOptions.MaxRetryTime()))
	if err != nil {
		return err
	}
	return nil
}
```

This reduces the lock time to only long enough to check `w.batching` (if batching is disabled) or to modify/extract the current batch (if batching is enabled).

Double-checked locking could also be used in tandem with `"sync/atomic".Bool` (from Go 1.19) which would avoid the lock entirely if batching is disabled.

__Alternatives considered:__

- Use the asynchronous API instead: this complicates error handling. I think the async API is great for recording app metrics but not so much for batch writing offline time-series data to InfluxDB.
- Use `api.NewWriteAPIBlocking` instead of `Client.WriteAPIBlocking` to purposely create multiple `writeAPIBlocking` instances, one per goroutine: this seems to go against the grain of the API and also creates many redundant `Service`s as well
- Use Go's standard HTTP client and only rely on InfluxDB client for line protocol formatting: this is certainly a possibility but is a lot more work

__Use case:__

- Throughput when batch writing: the server can handle simultaneous writes to the same bucket from many sources, but the client bottlenecks to a single operation at a time

vlastahajek:
Closes #352 

## Proposed Changes

More efficient synchronization around batching related code in WriteAPIBlocking

## Checklist

- [X] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/api/writeAPIBlocking.go
+++ b/api/writeAPIBlocking.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	http2 "github.com/influxdata/influxdb-client-go/v2/api/http"
 	"github.com/influxdata/influxdb-client-go/v2/api/write"
@@ -51,9 +52,10 @@ type WriteAPIBlocking interface {
 type writeAPIBlocking struct {
 	service      *iwrite.Service
 	writeOptions *write.Options
-	batching     bool
-	batch        []string
-	mu           sync.Mutex
+	// more appropriate Bool type from sync/atomic cannot be used because it is available since go 1.19
+	batching int32
+	batch    []string
+	mu       sync.Mutex
 }
 
 // NewWriteAPIBlocking creates new instance of blocking write client for writing data to bucket belonging to org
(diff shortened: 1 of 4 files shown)


api/writeAPIBlocking.go
```diff
@@ -69,28 +70,28 @@ func NewWriteAPIBlockingWithBatching(org string, bucket string, service http2.Se
...
+		return func(b string) error {
+			w.mu.Lock()
+			defer w.mu.Unlock()
+			w.batch = append(w.batch, b)
+			if len(w.batch) == int(w.writeOptions.BatchSize()) {
+				return w.flush(ctx)
+			} else {
+				return nil
+			}
+		}(line)
```
sranka, line 92: Is there a reason to wrap the code into a function that is immediately invoked?

api/writeAPIBlocking.go
```diff
@@ -69,28 +70,28 @@ func NewWriteAPIBlockingWithBatching(org string, bucket string, service http2.Se
 }
 
 func (w *writeAPIBlocking) EnableBatching() {
-	w.mu.Lock()
-	defer w.mu.Unlock()
-	if !w.batching {
-		w.batching = true
+	if atomic.LoadInt32(&w.batching) == 0 {
```
sranka, line 73: Wouldn't be easier to enforce or document that EnableBatching can be called before any write is performed?

sranka (changes requested):

api/writeAPIBlocking.go
vlastahajek: The reason is the lock will be on during call flush and will be unlocked after this. The lock doesn't need to be on later.

sranka: ... there is IMHO no 'later' in the `write` function, the result of the function is directly returned

api/writeAPIBlocking.go
vlastahajek: It would be easier to document it. I don't think it will be easier to enforce it. The way how it is safe.

api/writeAPIBlocking.go
vlastahajek: You are right, the function doesn't need to be there. Even if the func wrapper is removed, both returns will be there, and unlock will be called immediately. I will remove it.

PR #358, opened 2022-10-12
feat: Allow setting of application name to the User-Agent header value

### LINKED ISSUE #356: Allow overriding the default User-Agent header
__Proposal:__
Add a feature allowing users of this client library to override the default `influxdb-client-go` User-Agent HTTP header. This would make it easier to understand which apps are making which requests. E.g., currently, if App1 and App2 use this client, the UA coming from both will read `influxdb-client-go`. Allowing the apps to override the UA would make it easier to distinguish between the two.

__Current behavior:__
User-Agent header is hard coded to `influxdb-client-go`

__Desired behavior:__
Add 2 options for users:
1. Add method to set a field or two (AppName and Version?) for their User-Agent. The client would then construct a User-Agent. Something like `influxdb-client-go/2.3 AppName/Version`. Mark this in the docs as **recommended**.
2. Add a second method that allows completely overriding the User-Agent.

vlastahajek:
Closes #356 

## Proposed Changes

Added new option for setting the application name. Application name is then added to the end of Go client version string: E.g. `Influxdb-client-go/2.10.0 (linux; amd64) IoTHub/1.10`
- Added also example how to completely overwrite User-Agent header value using the Doer interface 

## Checklist

- [X] CHANGELOG.md updated
- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/client.go
+++ b/client.go
@@ -145,6 +145,9 @@ func NewClientWithOptions(serverURL string, authToken string, options *Options) 
 		}
 		ilog.Infof("Using URL '%s'%s", serverURL, tokenStr)
 	}
+	if options.ApplicationName() == "" {
+		ilog.Error("Application name is not set")
+	}
 	return client
 }
 
--- a/client.go
+++ b/client.go
@@ -146,7 +146,7 @@ func NewClientWithOptions(serverURL string, authToken string, options *Options) 
 		ilog.Infof("Using URL '%s'%s", serverURL, tokenStr)
 	}
 	if options.ApplicationName() == "" {
-		ilog.Error("Application name is not set")
+		ilog.Warn("Application name is not set")
 	}
 	return client
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
 ## [unreleased]
+### Features
+- [#358](https://github.com/influxdata/influxdb-client-go/pull/358):
+    - Added possibility to set an application name, which will be part of the User-Agent HTTP header
+        - Set using `Options.SetApplicationName`
+        - Error message is written to log if an application name is not set.
+    - Added example how to fully override `User-Agent` header using `Doer` interface
+
 ### Bug fixes
 - [#359](https://github.com/influxdata/influxdb-client-go/pull/359) `WriteAPIBlocking.Flush()` correctly returns nil error. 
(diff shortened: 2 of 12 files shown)


client.go
```diff
@@ -145,6 +145,9 @@ func NewClientWithOptions(serverURL string, authToken string, options *Options)
 		}
 		ilog.Infof("Using URL '%s'%s", serverURL, tokenStr)
 	}
+	if options.ApplicationName() == "" {
+		ilog.Error("Application name is not set")
```
sranka, line 149: it should be a warning, and written at most once per application run

sranka (changes requested):

client.go
sranka: > As the application name is an option per client instance, the `NewClient...` function is called once per client instance lifecycle.

ok

> Error level is chosen because it is the default log level. I agree a warning would be better, but warning will not be shown by default.
Personally, I think that `NewClient` should be also deprecated (or maybe reorganized), now it will always cause an error to be logged. Please add other approvers and let them decide.

vlastahajek:
@lesam, As you noted in issue #356, there should be an annoying log message, if the application name is unset. 
The current state of the logging mechanism is that only errors are shown by default. Thus writing such a message which would be always visible in the log requires writing it as an error.
But, it will introduce this annoying message to all users. This doesn't look like friendly instruction for a new feature. Writing it as a warning seems to be better.
WDYT?

lesam:
It seems reasonable to start with a warning, put it in the release notes/docs, and then maybe upgrade it to an error message in the next version?

@dgnorton for thoughts as well.

lesam:
Just leaving it as a warning forever is not sufficient, since it will never be noticed.

vlastahajek:
@lesam, @dgnorton, @powersj, the release is just outside the door, we have to decide if to stay with the error level or change the message to warning.

vlastahajek:
We had an internal discussion in Bonitoo and we are in favor of a warning because such a small feature should not bug users.

bednar (commented):
I prefer to use `warning` and add a warning to `CHANGELOG.md` that it will change to `error` in a future version of the client.

bednar (approved):
LGTM 🚀

PR #370, opened 2023-01-12
feat: add the ability to query with SQL

powersj:
The latest version of InfluxDB Cloud comes with the ability to send SQL queries. This is done using a new gRPC+TLS client. For now the basic query function will create the different, non-REST client and return an Arrow record slice. Users can then go through that slice as necessary.

- [ ] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x] Tests pass


README.md
```diff
@@ -427,6 +428,50 @@ func main() {
...
+
+func main() {
+	client := influxdb2.NewClient("http://us-east-1-1.aws.cloud2.influxdata.com:8086", "my-token")
+    defer client.Close()
+
+	querySQLAPI := client.QuerySQLAPI()
+	result, err := querySQLAPI.Query(context.Background(), "my-bucket", "select * from cpu limit 10")
+	if err != nil {
+		panic(err)
+	}
```
srebhan, line 458: Indentation seems to be messed up here...

README.md
```diff
@@ -427,6 +428,50 @@ func main() {
...
+    for {
+		if nextRecord, err := reader.Read(); err == io.EOF {
+			reader.Release()
+			break
+		} else if err != nil {
+			panic(fmt.Errorf("unable to read record: %v", err))
+		} else {
+			fmt.Print(nextRecord)
+		}
+	}
```
srebhan, line 469: I don't see `reader` defined anywhere, did you mean `result` here?

srebhan (commented):
Looks mostly good, only have two comments on the example.

README.md
```diff
@@ -427,6 +428,50 @@ func main() {
...
 }
 ```
+
+### SQL
+
+InfluxDB Cloud now supports the ability to query with SQL. With this support
+a user can pass in an SQL query to get a Arrow reader and iterate through the
+records.
+
+> :warning: Parameterized Queries are supported only in InfluxDB Cloud. There is no support in InfluxDB OSS currently.
```
vlastahajek, line 438: The warning should probably say that SQL support is in the cloud only

README.md
vlastahajek, line 469: I believe this is completely wrong. The general Go idiom for iterating over result set is using `obj.Next()` and check for an error via `obj.Err()`, an the retuned reader supports it:
```go
    defer result.Release()
    for result.Next() {
        fmt.Println(result.Record())
    }
    if result.Err() != nil {
        panic(fmt.Errorf("unable to read record: %v", err))
    }
```

vlastahajek: Also, such an example needs to be added to API docs examples in `api/examples_test.go`. 
Note: to overcome the circular import issue there is a fake client that needs to be enhanced with the QuerySQLAPI() function.
```go
func ExampleQuerySQLAPI() {
	client := influxdb2.NewClient("http://us-east-1-1.aws.cloud2.influxdata.com:8086", "my-token")
	defer client.Close()
	querySQLAPI := client.QuerySQLAPI()
	result, err := querySQLAPI.Query(context.Background(), "my-bucket", "select * from cpu limit 10")
	if err != nil {
		panic(err)
	}
	defer result.Release()
	for result.Next() {
		fmt.Println(result.Record())
	}
	if result.Err() != nil {
		panic(fmt.Errorf("unable to read record: %v", err))
	}
}
```

README.md
srebhan: Hmmm not sure if this is a cloud thing, but rather an IOx vs. InfluxDBv2 issue...

powersj:
Thanks for the review, however I will close this PR.

As-is this PR does not provide anything unique or special over using the Arrow FlightSQL library directly. SQL query support is fundamentally a different animal compared to the existing library and REST support that it provides.

PR #389, opened 2023-10-17
feat: Allow struct field tags for influxdb with a marshaller

zob456:
Closes #

## Proposed Changes

My suggested change creates a function in `api` called `MarshalStructToWritePoint` that will take in a value of a custom type & return either a `*write.Point` correctly formatted or an `error`

The company I work for works with Influx & I've had to write a custom marshaller for just this reason. I also found some other threads on stackoverflow & other places asking if this was on the road map / covered by another package.

I haven't contributed to a go package like this before so I hope I am in-line with best practices here.

Any feedback is greatly appreciated even if my PR is not accepted!

Thanks!

## Checklist

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergeable
- [x] Tests pass
- [ ] Sign [CLA](https://www.influxdata.com/legal/cla/) (if not already signed)


powersj:
Thanks for the PR! I'll add this to our list to review and discuss the feature addition next week.

Thanks again!

powersj:
@zob456,

Thanks for the PR. I think we would be interested in landing this. One idea we thought about is it would be nice to get a unmarshaller as well, so you could do the full loop.

I'll add this to the backlog to get you a full, proper review.

zob456:
@powersj great point! I hadn't thought of an Unmarshaller but that would be very helpful!

I'll look out for the backlog issue!

Thanks again!

zob456:
@powersj wanted to check back in & see if this is tracked somewhere as an issue so I can work on the additional functionality & open a proper PR?

powersj:
Hi @zob456,

Ah I have the review of this PR tracked in the backlog for someone to provide a proper review. I do not have a public issue for the unmsarshaller, but feel free to put something up. Hopefully that answers your question.

Thanks

sranka (commented):
Thank you for your PR.  There are a few nits in the implementation, but it looks good. The primary question is: "Should this be part of the client library?". 

With the functionality introduced herein, the client library would internally
1. convert a struct to `write.Point` with a help of `influxdb` tag on struct fields (scope of this PR)
2. convert [Point to lp.Metric](https://github.com/influxdata/influxdb-client-go/blob/0f9ebbed21d23ce6850ba75d2a0bc8feb730f104/internal/write/service.go#L198)
3. encode lp metric to a protocol line and write protocol line(s)

I think that the conversion to a Point instance introduced herein is redundant ... the client library should better introduce a new `WriteData` function on `WriteAPI` as it was planned for a v3 version of this library. I would also prefer the approach taken in https://github.com/influxdata/influxdb-client-go/blob/91f06dfd4928dd6e95b123dc5b072abebc0df208/influxclient/writer.go#L152 , including annotations names and structure.

I think that this PR can introduce a good utility function outside the client library, but it should rather not become a new enhancement of the library.

sranka:
I would be fine with the tags and the overall approach that already went through review in a v3 client in https://github.com/influxdata/influxdb-client-go/blob/91f06dfd4928dd6e95b123dc5b072abebc0df208/influxclient/writer.go#L152.

sranka:
https://github.com/influxdata/influxdb-client-go/pull/394 adds the same functionality using struct tags that were planned for v3 version of this library

